Issue #16, opened 2017-01-15
Cannot remap_keys when using "window focus follows the mouse"


acro5piano:
Thanks for the life-changing tool! 

I am using [awesome wm](https://awesomewm.org).
Awesome wm use "window focus follows the mouse" feature. 
The feature enables to change the window focus without clicking windows.

I found xremap cannot remap keys if the window focus was changed by the mouse focus.
I read the source code and I found adding the following line solve this problem.

```diff
--- a/mrblib/xremap/active_window.rb
+++ b/mrblib/xremap/active_window.rb
@@ -10,6 +10,7 @@ module Xremap
     end

     def changed?
+      sleep 0.1
       next_window = fetch_active_window
       @current_window != next_window
     ensure
```

However, I don't think this is the best solution...

Maybe `Xremap::ActiveWindow#fetch_active_window` fetches faster than the mouse focus.
So the problem will be fixed if `XlibWrapper.fetch_active_window` (in xlib_wrapper.c) wait until the active window change, or adding delay.

But I cannot imagine how to do it.
I am going to create a pull request soon.

Thanks

PR #17, opened 2017-01-15
Fix cannot remap_keys when using "window focus follows the mouse"

acro5piano:
Thanks for the life-changing tool!

I am using [awesome wm](https://awesomewm.org).
Awesome wm use "window focus follows the mouse" feature. 
This feature enables to change the window focus without clicking windows.

As I mentioned in #16, I found xremap cannot remap keys if the window focus was changed by the mouse focus.

This patch fixes the problem adding delay before `fetch_active_window`. 

Thanks

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
 /xremap
 /mruby
+tags
+TAGS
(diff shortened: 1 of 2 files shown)


k0kubun:
Oh, it's a surprising fact. Ideally we should hook an *actual* focus change event because it delays remapping and makes usability worse. But for now, accepting the patch. Thank you for your help!

acro5piano:
Thank you!

Issue #15, opened 2017-01-14
Provide a way to remap the case "remapped for class: false"


k0kubun:
In some situations (like https://github.com/k0kubun/xremap/issues/13), xremap says "remapped for class: false". It may be difficult to remap keys in windows except it.

So we should do following things.

- Research what kind of window has no class and write it as limitation.
- Provide another way to distinguish windows. At least we can know pid and cmdline of it.
  - In that case, xremap should log these information.

acro5piano:
In my environment, the following applications have no class.

- Emacs
- Terminator
- eog
- Firefox developer edition
- pcmanfm

Also the following applications have a class.

- vlc
- chromium
- urxvt

Thanks

PR #19, opened 2017-05-05
Make fetch_window_class look up the parent window if get WM_CLASS failed

hanachin:
Some applications have the window without name, such as Firefox.

c.f.
> XGetInputFocus() returns an unnamed child window
> that every instance of Firefox has.
https://burrows.svbtle.com/universal-copy-paste-in-linux

### DIFF
--- a/src/xlib_wrapper.c
+++ b/src/xlib_wrapper.c
@@ -17,7 +17,18 @@ mrb_xw_fetch_window_class(mrb_state *mrb, mrb_value self)
   Atom net_wm_name = XInternAtom(display, "WM_CLASS", True);
 
   XTextProperty prop;
-  XGetTextProperty(display, window, &prop, net_wm_name);
+
+ get_wm_name:
+
+  if (!XGetTextProperty(display, window, &prop, net_wm_name)) {
+    unsigned int nchildren;
+    Window root, parent, *children;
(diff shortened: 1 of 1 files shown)


hanachin:
Fix #15 due to get WM_CLASS correctly.

k0kubun:
Oh cool! 

For now I can't test your feature, so I'll test and merge this after I go back to my house with Linux desktop.

src/xlib_wrapper.c
```diff
@@ -17,7 +17,18 @@ mrb_xw_fetch_window_class(mrb_state *mrb, mrb_value self)
...
+
+ get_wm_name:
+
+  if (!XGetTextProperty(display, window, &prop, net_wm_name)) {
+    unsigned int nchildren;
+    Window root, parent, *children;
+    if (XQueryTree(display, window, &root, &parent, &children, &nchildren)) {
+      if (children) { XFree(children); }
+      window = parent;
+      goto get_wm_name;
```
k0kubun: Is this `goto` inevitable? Can't we use infinite loop and break?

hanachin: @k0kubun It's evitable, I changed `goto` to infinite loop at a9d8c0c . Could you please test again?

k0kubun:
Working fine. Thanks!

PR #25, opened 2018-01-03
Fix event handling lag

lostdj:
On KDE 4 switching between virtual desktops results in a flood of `PropertyNotify` event: https://github.com/k0kubun/xremap/blob/cef2ca36ee0392f1ac2cd775c200c721f798dd2f/tools/xremap/main.c#L56

Coupled with the specified delay of 100ms in this file, and the fact that `KeyPress` event is handled last, it results in a 1+ second lag.

I don't know whether there's a justification for the existing, specific 100ms delay, but 1ms seem to be ok for me.

### DIFF
--- a/mrblib/xremap/active_window.rb
+++ b/mrblib/xremap/active_window.rb
@@ -19,7 +19,7 @@ module Xremap
     private
 
     def fetch_active_window
-      sleep 0.1
+      sleep 0.001
       XlibWrapper.fetch_active_window(@display)
     end
   end


k0kubun:
Thanks, let's try this delay.

acro5piano:
Hi there,

I found the 1ms delay is not ok for my environment(Awesome WM).

```
$ awesome --version
awesome v3.5.6 (For Those About To Rock)
 • Build: Sep  1 2015 11:28:53 for x86_64 by gcc version 5.2.1 (buildd@lgw01-18)
 • Compiled against Lua 5.1.5 (running with Lua 5.1)
 • D-Bus support: ✔

$ uname -a
Linux kazuya-ThinkPad-X240 4.4.0-109-generic #132-Ubuntu SMP Tue Jan 9 19:52:39 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

As @k0kubun mentioned in [this thread](https://github.com/k0kubun/xremap/pull/17) ideally xremap can handle the actual focus change.
Or users can set the delay time like `./configure --focus-delay=0.1` for the time being...

k0kubun:
Not tested yet, I commit a change for you https://github.com/k0kubun/xremap/commit/194e1d4be0638317af5fe32dd9fb758da768941a (also changed back default too). Could you try it and changing it via `XREMAP_DELAY=0.001 xremap`?

acro5piano:
@k0kubun 
Thank you for quick response!
I tested the feature and seems to be working! 

Thanks again this life changing tool!

PR #26, opened 2018-01-03
guard segfault if window not found

waka:
The segmentation fault error has occured.

screenshot from 2018-01-04 01-42-17

I use GNOME and wayland as default x server.
It seems that window can not be taken when focusing on GNOME application such as Nautilus.


k0kubun:
Thank you for fixing!

PR #30, opened 2018-09-05
Support for XFree86 vendor specific 

igui:
Add support for Multimedia keys, usually associated with the `Fn` key in some laptop keyboards and special keys in Internet Keyboards.

I added an example for it to test.


mrblib/xremap/key_expression.rb
```diff
@@ -38,7 +38,11 @@ def modifier_mask(modifiers)
       end
 
       def to_keysym(keyexp)
-        X11.const_get("XK_#{keyexp}")
+        begin
+          X11.const_get("XK_#{keyexp}")
+        rescue NameError # Allow non "XK_" expressions
```
k0kubun: Could you switch the behavior by `if keyexp.start_with?('XF86XK')` rather than branching by exception?

k0kubun:
Merged and applied the above by myself. Thank you for your patch.

PR #31, opened 2018-09-13
Generate X11 constants from headers at build-time.

hanachin:
The X11 constants are different between versions.

For example, Debian doesn't have XF86XK_WWAN
https://salsa.debian.org/xorg-team/proto/x11proto-core/blob/debian-unstable/XF86keysym.h

But Arch Linux does.
https://www.archlinux.org/packages/extra/any/xorgproto/

I've got following error on Debian stretch

```
 /home/sei/src/github.com/k0kubun/xremap/src/x11_constants.c:2710:20: error: ‘XF86XK_WWAN’ undeclared (first use in this function)
    define_x11_const(XF86XK_WWAN);
```

To avoid macro undeclared error, wrap the `define_x11_const` with #ifdef

```
#ifdef XF86XK_WWAN
define_x11_const(XF86XK_WWAN)
#endif
```

Or generate code at build-time.


k0kubun:
Wow, this looks cool :+1: As building mruby requires ruby, this wouldn't add extra build dependency.

k0kubun:
I invited you to collaborators of xremap.

Makefile
```diff
@@ -22,5 +22,11 @@ mruby:
 	git clone https://github.com/mruby/mruby
 	git -C mruby reset --hard $(REVISION)
 
-mruby/build/host/bin/xremap: mruby build_config.rb $(CSRCS) $(MRBSRCS) $(MRBCSRCS)
+src/x11_constants.inc:
+	cat /usr/include/X11/keysymdef.h | ruby -e 'puts STDIN.read.split("\n").select {|l| l.match(/\A(#define XK_|#ifdef|#endif)/) }.map{|l| l.match(/\A#define XK_/) ? %Q[  define_x11_const(#{l.split(" ")[1]});] : l }.join("\n")' > src/x11_constants.inc
+
+src/x11_constants_XF86.inc:
+	cat /usr/include/X11/X.h | ruby -e 'puts STDIN.read.split("\n").select {|l| l.start_with?("#")}[2..-2].map{|l| l.start_with?("#define") ? %Q[  define_x11_const(#{l.split(" ")[1]});] : l}' > src/x11_constants_XF86.inc
```
hanachin: Ooops, this should be `cat /usr/include/X11/XF86keysym.h`, I'll fix it.

k0kubun:
Hi. As you might have noticed, we replaced the implementation and it's now Rust-based. Now I'm thinking about automating releases by setting GitHub Secrets to be used by GitHub Actions, but credential management would be complicated if the repository's write access is shared.

As a workaround for that problem, assuming you may not be interested in actively maintaining the current version, let me remove you from collaborators for now. However, if you're interested, I'm 100% open to welcoming you as a collaborator again. So please feel free to let me know if that's the case. I'll think about other ways to automate releases in case it's needed.

PR #36, opened 2020-04-18
Pass unmatched modifiers

genya0407:
This patch implements "pass-through" of unmached modifiers.

## Why

Karabinar ignores unmached modifiers, and apply ignored modifiers to modified key input.
For example, if we map `Ctrl+j` to `↓`(keydown), inputs like `Ctrl+Shift+j` is modified to `Shift+↓` so we can select multiple lines in text editors.

On the other hand, xremap does not ignore unmached modifiers (in previous example, `Shift`), so `Ctrl+Shift+j` maches to nothing, and we cannot select lines.

## What

With this patch, xremap maches modifiers to their "superset" modifiers, and apply remaining modifiers to the resulting key inputs. For example, if `Ctrl+Shift+j` is input, it matches to `Ctrl+j`, and apply `Shift` to resulting key input `↓`.

### DIFF
--- a/mrblib/xremap/event_handler.rb
+++ b/mrblib/xremap/event_handler.rb
@@ -3,18 +3,19 @@ module Xremap
     # @param [Xremap::Config] config
     # @param [Xremap::Display] display
     def initialize(config, display)
+      @display = display
       @active_window = ActiveWindow.new(display)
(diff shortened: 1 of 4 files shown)


genya0407:
Sorry, this patch has a bug. I'll fix it.

PR #51, opened 2021-12-30
Add hotkey command support

gdvalle:
Hello,

I was wondering if you might be interested in an extension of xremap to support something like [xbindkeys](https://www.nongnu.org/xbindkeys/), where a key can be mapped to an external command.  I had a similar need you described, where app-specific bindings were useful, and I thought it could be a little cleaner than having that in the config rather than a window check in each command. I found your project here and was able to hack this in quickly!

What's here so far is pretty simple. A new config kind is added, `hotkeys`, which you can specify the same `application` and `name`, and `keys` for each hotkey to trigger from. `command` for specifying the executable and arguments as an array.  As implemented it happens _after_ remap, but before the keymap phase.

A few thoughts where this could go:
* Add a `forward_key_press` option. As-is it swallows the key press, but it could propagate that on.
* Wire this up such that it happens after keymap. Or go backward, and do it before remap. I thought it made more sense to do after remap.
* Build a table of hotkey commands at config parse, so we don't have to iterate every command (performance).
* Docs, examples.

Thanks for taking a look, and I fully understand if this is not something you're interested in supporting.

Thanks for the great project!

### DIFF
--- a/example/hotkey_test.yml
+++ b/example/hotkey_test.yml
@@ -1,0 +1,10 @@
+hotkeys:
+  - name: hotkey_test
+    keys:
+      - KEY_GRAVE
+    application:
+      not: [Google-chrome, konsole]
+    command:
+      - "/bin/sh"
+      - "-c"
+      - "date > /tmp/hotkey_test"
--- a/src/command.rs
+++ b/src/command.rs
@@ -1,0 +1,32 @@
+use std::process::{Command, Stdio};
+use std::thread;
+use log::{debug, error};
+
+pub fn run_command(command: Vec<String>) {
+    // To avoid defunct processes, spawn a thread to wait on the process.
+    thread::spawn(move || {
+        debug!("Running command: {:?}", command);
+        match Command::new(&command[0])
+            .args(&command[1..])
+            .stdin(Stdio::null())
+            .stdout(Stdio::null())
+            .stderr(Stdio::null())
+            .spawn()
+        {
+            Err(e) => {
+                error!("Error running command: {:?}", e);
+            }
+            Ok(mut child) => {
+                debug!("Process spawned: {:?}, pid {}", command, child.id());
+                match child.wait() {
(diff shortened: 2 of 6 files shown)


k0kubun:
Thank you! The YAML syntax was designed with a feature like this in mind, so I can support this issue without any issue. The idea is to support it in `keymap` as an element of `Action`. Since I'm working on simplifying the config serde now, let me take over your work after that.

k0kubun:
@gdvalle Could you test this branch https://github.com/k0kubun/xremap/pull/52? I changed the syntax a bit. I'll think about changing the implementation to wait for child processes using a `SIGCHLD` handler instead of making a thread for each child process, but the current behavior will remain.

gdvalle:
Thanks, @k0kubun! #52 works great. Makes much more sense to extend the Action behavior like that, and having it in keymap.

`SIGCHLD` handler could be a good answer for avoiding the extra threads. Another option, platform-specific, but probably fine in this case, is to set the `SA_NOCLDWAIT` flag on a handler. I tried this using the `nix` crate and it seems to work correctly on Linux:
```rust
use nix::sys::signal::{SaFlags, SigAction, SigSet, SigHandler, Signal, sigaction};
let sig_action = SigAction::new(SigHandler::SigDfl, SaFlags::SA_NOCLDWAIT, SigSet::empty());
unsafe { sigaction(Signal::SIGCHLD, &sig_action).expect("Failed to register SIGCHLD handler"); }
```
Here's an example using libc directly: https://github.com/cube-js/cube.js/blob/9945b0b570423b0d974dcfff77472912710cdb98/rust/cubestore/cubestore/src/sys/process.rs#L1-L19

I'll go ahead and close this in favor of #52.

k0kubun:
Good call. I stole your idea. Released this feature as v0.1.6.

gdvalle:
Thanks @k0kubun! Much appreciated!

PR #63, opened 2022-01-11
support more mouse buttons

KeenS:
amends #58 .
See also https://github.com/torvalds/linux/blob/7b229b13d78d112e2c5d4a60a3c6f602289959fa/include/uapi/linux/input-event-codes.h#L342-L362


k0kubun (approved):
Thank you!

Issue #40, opened 2021-12-22
[keymap] application only filter does not work on Sway


acro5piano:
First of all, thank you for the great library! I'm a big fan of Xremap and using it for over five years. It significantly improves my productivity in my daily Linux life!

# Problem

application only filter seems not work on Sway WM.

- expected: Xremap is applied to specific windows
- actual: Xremap is not applied to any windows

# Versions

- Xremap 0.1.0
- `Linux archlinux 5.15.7-arch1-1 #1 SMP PREEMPT Wed, 08 Dec 2021 14:33:16 +0000 x86_64 GNU/Linux`
- sway version 1.6.1

# Config

```yml
keymap:
  - name: Emacs binding
    application:
      only: 
        - brave-browser # wayland Brave
        - Brave-browser # xwayland Brave
        - chromium      # wayland Chromium
    remap:
      C-b: left
      C-f: right
      C-p: up
      C-n: down
```

And run

```
sudo ./Downloads/xremap .xremap.yml --device /dev/input/event4
```

(I disabled all other software like xremap(ruby), wayremap, and XKeysnail)

# What I tried

### To see whether xremap supports Xwayland or not, launch Brave in Walyand mode

```
brave --enable-features=UseOzonePlatform --ozone-platform=wayland
```

Result: Not works

### How about Chromium ?

```
chromium --enable-features=UseOzonePlatform --ozone-platform=wayland
```

Result: Not works

# Other notes

To see Window Class and App ID in Sway, the following command is useful:

```
swaymsg --type subscribe --monitor '["window"]' | jq -r '.. | select(.focused?) | [.app_id, .window_properties.class] | @tsv'
```

k0kubun:
Thank you for filing this. May I first check the following things?

* With that config, once you press any of `C-b`, `C-f`, `C-p`, or `C-n`, xremap should print a log like `application-client: {} (supported: {})`. What does it say?
* How did you install the binary? Is it a binary downloaded from Releases, or did you build it? If you haven't tried it yet, could you try building xremap with `cargo build xremap --release --features sway` using `target/release/xremap`?

acro5piano:
I tried again with Chromium with native Wayland it worked. At that time Xremap logs `supported: true` but I couldn't reproduce it for now. I'll try again later.

k0kubun:
I'm glad that it could work. Let me close this until this issue becomes actionable on my end, but please reopen this if you figure out how to reproduce it.

[...]

PR #65, opened 2022-01-22
Handle sway application support for XWayland windows

markk:
XWayland windows have null app_id, look for and return window_properties.class instead.

Perhaps relevant to this issue https://github.com/k0kubun/xremap/issues/40


k0kubun (approved):
Oh, interesting. Thanks! FYI @acro5piano

PR #68, opened 2022-01-31
Only use feature-related dependencies when needed.

cireu:
This help xremap only pull x11_rs/swayipc/zbus only when corresponding feature flags is used.


k0kubun (approved):
Thank you!

PR #77, opened 2022-02-16
move command line parsing to clap; address clippy

ModProg:
On top of switching to clap, I did some clippy fixes, if you disagree with any I'm happy to roll them back.

I also added support for generating shell completions


k0kubun (approved):
:clap:

PR #78, opened 2022-02-22
Device Watcher improvements and Config reload

ModProg:
## Fix Device Watching
- new struct `InputDevice` to hold both the path and the `Device`
- support paritial device loading when watching, so that only new devices get registered
- fix logic of device watcher

## General
- avoid regenerating watchers by moving them outside the event loop

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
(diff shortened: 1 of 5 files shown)


k0kubun:
Could you summarize a list of changes in this PR in your PR description?

k0kubun (approved):
Thanks!

PR #82, opened 2022-04-01
Add timeout_millis for nested remap

k0kubun:
## Example
```yml
keymap:
  - remap:
      C-x:
        remap:
          a: b
        timeout_millis: 1000
```

* C-x -> 500ms -> a
  * b
* C-x -> 1500ms -> a
  * a

### DIFF
--- a/src/config/action.rs
+++ b/src/config/action.rs
@@ -1,9 +1,11 @@
 use crate::config::key_press::KeyPress;
 use std::collections::HashMap;
 
+use crate::config::remap::Remap;
 use serde::de;
 use serde::{Deserialize, Deserializer};
 use std::fmt::Debug;
+use std::time::Duration;
 
 // Values in `keymap.remap`
 #[derive(Clone, Debug, Deserialize)]
@@ -11,7 +13,7 @@ use std::fmt::Debug;
 pub enum Action {
     KeyPress(KeyPress),
     #[serde(deserialize_with = "deserialize_remap")]
-    Remap(HashMap<KeyPress, Vec<Action>>),
+    Remap(Remap),
     #[serde(deserialize_with = "deserialize_launch")]
     Launch(Vec<String>),
     #[serde(deserialize_with = "deserialize_set_mark")]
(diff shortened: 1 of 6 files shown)


0x78logs:
that's wonderful. just finished the build and did a couple of tests

here's the config that I have used:

```
keymap:
  - remap:
      j:
        remap:
          k: Esc
        timeout_millis: 200
```

the issue is if I click `j` alone it will never resolve. it stays there... so it makes `j` unusable in my keyboard

another note: it would very nice if we can control the default 500ms.

k0kubun:
> another note: it would be very nice if we can control the default 500ms.

For backward compatibility, when it doesn't have a timeout, it doesn't timeout. I think it's natural from how it's configured, too.

0x78logs:
but I can't assign a timeout to the first `remap`
I can't do something as the following:

```
keymap:
  - remap:
      j:
        remap:
          k: Esc
    timeout_millis: 200
```

Error: "mapping values are not allowed in the context ..."

k0kubun:
It's top-level, there's nothing to be timed out. All you need is  https://github.com/k0kubun/xremap/pull/82#issuecomment-1085834567.

0x78logs:
I don't understand, it doesn't work ... pressing `j` and not pressing the following key in the sequence won't resolve to the normal behaviour of `j` thus making `j` in your keyboard not functional.

that's not the behavior of what this feature should achieve.

k0kubun:
In case I confused you by replying a short meesage,

> the issue is if I click j alone it will never resolve. it stays there... so it makes j unusable in my keyboard

I still intend to fix this. However, it doesn't still change the fact that the correct config is:

```yml
keymap:
  - remap:
      j:
        remap:
          k: Esc
        timeout_millis: 200
```

because of how the indent level impacts which key. `timeout_millis` in the above YAML timeouts `j`, so if you're interested in the timeout of `j`, that's what you need.

k0kubun:
I updated the implementation. This YAML:

```yml
keymap:
  - remap:
      j:
        remap:
          k: Esc
        timeout_millis: 200
```

behaves like:

* j -> 100ms -> k
  * Esc
* j -> 300ms -> k
  * j -> k
* j -> any ms -> l
  * j -> l

0x78logs:
Hello, the config is working as expected now.
however there's still the issue of `j` not resolving if it's not followed by another key.

for example
right now I cannot use my hjkl in vim because it will stuck on `j`

below is the only behavior that's missing 
j -> 200ms or more, user didn't follow with another key press ...
    j

Thank you.

k0kubun:
I updated the implementation. This YAML:

```yml
keymap:
  - remap:
      j:
        remap:
          k: Esc
        timeout_millis: 200
```

behaves like:

* j -> 100ms -> k
  * Esc
* j -> 300ms -> k
  * j -> k
* j -> any ms -> l
  * j -> l
* j -> 300ms
  * j

0x78logs:
I'm running on the new implementation, so far it is working great.

Give me a couple of days to test it and I'll report if something
misbehaves.

Thank you so much!

k0kubun:
Maybe it's time to release this?

0x78logs:
Yes, just merge the branch. I'm very happy with the results.
Thank you.

k0kubun:
Thank you for your patience and testing.

PR #86, opened 2022-04-08
Add --daemon option

k0kubun:
## Example

```
$ xremap config.yml --daemon
```


k0kubun:
See https://www.reddit.com/r/commandline/comments/ttiefi/this_takashi_person_may_as_well_have_split_the/ for contexts.

PR #93, opened 2022-05-07
Support modal remapping

### LINKED ISSUE #89: Vim-like modal editing
First let me say, thanks for this project. I love it!

Recently I switched to evil-mode in emacs and I wondered if it would be (sort of) easy to add modal editing to xremap? Like a per-app modal state that switches keymaps according to mode. E.g. insert and visual mode.

That would be awesome in gtk apps, browser apps, and others that do not support modal editing.

I see that key combos are supported, this should allow something like `viw`-alike thingies.

k0kubun:
Close #89

## Example

```yml
default_mode: insert
keymap:
  - mode: insert
    remap:
      Esc: { set_mode: normal }
  - mode: normal
    remap:
      i: { set_mode: insert }
      h: Left
      j: Down
      k: Up
      l: Right
```

### DIFF
--- a/src/config/action.rs
+++ b/src/config/action.rs
@@ -16,6 +16,8 @@ pub enum Action {
     Remap(Remap),
     #[serde(deserialize_with = "deserialize_launch")]
     Launch(Vec<String>),
+    #[serde(deserialize_with = "deserialize_set_mode")]
+    SetMode(String),
     #[serde(deserialize_with = "deserialize_set_mark")]
     SetMark(bool),
     #[serde(deserialize_with = "deserialize_with_mark")]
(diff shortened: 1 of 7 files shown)


cipherlogs:
this is a terrific feature if it can be added, I have tested the code but it is not working at all

here are my bindings 

```
keymap:
  # - name: Global
  - mode: insert
    remap:
      j:
        remap:
          j: {set_mode: normal}
          k: Esc
          l: C-space
        timeout_millis: 100

  - mode: normal
    remap:
      i: {set_mode: insert}
      h: Left
      l: Right
      k: Up
      j: Down
```

to trigger normal mode I have to go `jj` but it doesn't work
also I have tried to add nested remaps inside normal mode it doesn't work at all

I hope this feature doesn't die. if we can do a true vim like modal remapping that would be amazing.

k0kubun:
Could you make it a separate issue so that I can remember it when I work on open issues?

PR #95, opened 2022-05-07
Support GNOME 41 and above without `unsafe_mode`.

wilfredwee:
Hello. First of all, thanks for this project. It's amazing and exactly what I'm looking for in Linux.

Since GNOME 41, gnome-shell has [locked down on internal API's](https://gitlab.gnome.org/GNOME/gnome-shell/-/merge_requests/1970), including
the `Eval` API. To avoid toggling on the `unsafe_mode` in GNOME, we
can instead install a gnome shell extension that provides a 'safe'
introspection API.

SafeIntrospection is a small gnome shell extension I created to expose
the
[Introspect](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/386d25e6f8ce11549526ea3776eb34138fcb3774/data/dbus-interfaces/org.gnome.Shell.Introspect.xml) API.

I have opened an issue with the GNOME Shell team to see if they can
provide this Introspect API publicly instead of having to rely on an
extension here: https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/5446

I have also updated the README to reflect the current status and new instructions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,43 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7a99269dff3bc004caa411f38845c20303f1e393ca2bd6581576fa3a7f59577d"
 
 [[package]]
+name = "async-broadcast"
+version = "0.4.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bbd92a9bd0e9c1298118ecf8a2f825e86b12c3ec9e411573e34aaf3a0c03cdd"
+dependencies = [
+ "easy-parallel",
+ "event-listener",
+ "futures-core",
+ "parking_lot",
+]
+
+[[package]]
+name = "async-channel"
+version = "1.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2114d64672151c0c5eaa5e131ec84a74f06e1e559830dabba01ca30605d66319"
+dependencies = [
+ "concurrent-queue",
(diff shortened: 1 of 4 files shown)


k0kubun:
Hi, thanks for looking into this. Coincidentally, I was working on this issue almost at the same time.

WDYT about https://github.com/k0kubun/xremap/pull/94? While it does require an extra step for installation, it's probably more sensible to avoid relying on internal API.

k0kubun:
Ah, sorry, it seems like yours also uses an extension when you don't want to use `unsafe`. So your patch is actually a superset of mine.

Anyway, if we rely on a third-party extension to make it work, I'd like to keep the full control under xremap's maintainers, so I'd like to still go with https://github.com/k0kubun/xremap/pull/94. When your proposal goes well, I'd be happy to use your implementation without `unsafe`, but meanwhile I'm thinking about just asking users to install an extension.

wilfredwee:
Hahah, interesting coincidence. Yeah that's fine. The extension I created is slightly more generic, but not a huge deal for this project.

k0kubun:
Thank you. When somebody files an issue about obviating extension installation, I might come back to this PR and cherry-pick your code, but for now, let me just go with https://github.com/k0kubun/xremap/pull/94.

I hope https://gitlab.gnome.org/GNOME/gnome-shell/-/issues/5446 goes well too!

k0kubun:
I initiated the release of my version as v0.4.0. Please let me know if you find any problems.

PR #98, opened 2022-05-17
Window matcher for modmap and keymap

k0kubun:
## Example
```yml
modmap:
  - remap:
      CapsLock: Esc
    window:
      only: Terminal
keymap:
  - remap:
      RO: Shift-RO
    window:
      not: Terminal
```


k0kubun:
* This should support a regexp.
* We should support X11 too. Maybe Sway as well if not too difficult.

PR #138, opened 2022-08-08
Support regex for application names

Frederick888:
I'd like to map left Ctrl as Esc when pressed alone unless I'm in
Minecraft. Since Minecraft adds its version number to the window class
name, e.g. `Minecraft* 1.19.2` (and I think I have an `*` cos I use
mods), it's nice to be able to use regular expressions.

Thanks for this great application btw!

### DIFF
--- a/src/config/application.rs
+++ b/src/config/application.rs
@@ -1,13 +1,91 @@
+use std::str::FromStr;
+
+use anyhow::anyhow;
+use regex::Regex;
 use serde::{Deserialize, Deserializer};
 
 // TODO: Use trait to allow only either `only` or `not`
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Application {
-    #[serde(default, deserialize_with = "deserialize_string_or_vec")]
-    pub only: Option<Vec<String>>,
-    #[serde(default, deserialize_with = "deserialize_string_or_vec")]
-    pub not: Option<Vec<String>>,
+    #[serde(default, deserialize_with = "deserialize_matchers")]
+    pub only: Option<Vec<ApplicationMatcher>>,
+    #[serde(default, deserialize_with = "deserialize_matchers")]
+    pub not: Option<Vec<ApplicationMatcher>>,
+}
+
+#[derive(Debug)]
+pub enum ApplicationMatcher {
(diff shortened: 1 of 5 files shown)


src/config/application.rs
```diff
@@ -1,13 +1,63 @@
...
+
+impl FromStr for ApplicationMatcher {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let first_char = s.chars().next();
+        match first_char {
+            None => Err(anyhow!("Empty application name")),
+            Some('~') if s.len() == 1 => Err(anyhow!("Empty regular expression for application name")),
+            Some('~') => Ok(ApplicationMatcher::Regex(Regex::new(&s[1..])?)),
```
k0kubun, line 40: I'm not used to the syntax of declaring a regular expression by starting it with `~`. How about wrapping a regex with `/`s [like CircleCI](https://circleci.com/docs/workflows#using-regular-expressions-to-filter-tags-and-branches), e.g. `/regex/`?

Also, it'd be nice to test parsing it at `src/config/tests.rs`.

Frederick888: Thanks for the review! Just two quick questions before I move forward:

1. Is there a straightforward way to unescape PCRE regular expressions and create `regex::Regex`? I'm happy to write a simple unescape function but just want to confirm I'm not missing anything
2. Do you also want it to support modifiers, e.g. `/^minecraft/i` for case-insensitive matching?

k0kubun: > Is there a straightforward way to unescape PCRE regular expressions and create regex::Regex? I'm happy to write a simple unescape function but just want to confirm I'm not missing anything

I don't know of one. I was just thinking about checking if it starts and ends with a `/` and removing them.

> Do you also want it to support modifiers, e.g. `/^minecraft/i` for case-insensitive matching?

No, I don't. For simplicity, I prefer starting from supporting no flags. I don't think the application name would fluctuate like minecraft, MineCraft, or Minecraft either.

Thanks for asking :slightly_smiling_face:

Frederick888: Pushed. Could you have a look again?

k0kubun (approved):
Thank you!

PR #147, opened 2022-08-14
Replace modifier: true with virtual_modifiers

k0kubun:
Replace https://github.com/k0kubun/xremap/pull/146 with the following syntax.

## Example
```yml
virtual_modifiers:
  - CapsLock
modmap:
  - remap:
      CapsLock:
        held: CapsLock
        alone: Esc
keymap:
  - name: Global
    remap:
      CapsLock-i: Up
      CapsLock-j: Left
      CapsLock-k: Down
      CapsLock-l: Right
```


ngirard:
Great ! Keep up the nice work !

PR #177, opened 2022-09-20
Add TAG+="uaccess" in the udev's rule

dottorblaster:
As above, on OpenSUSE the udev rule doesn't work unless one appends `TAG+="uaccess"` that basically means granting access to `uinput` to logged in users.


k0kubun (approved):
Thanks!

Issue #180, opened 2022-09-27
Support for scrollwheel events


timbertson:
I did a little hacking locally to try and add scrollwheel events. Specifically I'd like to switch tabs on alt+scrollwheel, so I only want to remap scroll events, not produce them.

My basic idea is that there are already SCROLLUP and SCROLLDOWN keys. For the purpose of config, it seems convenient to (re)use these for mapping mouse wheel events.

The actual events are different, they come through as RelAxis(REL_WHEEL) events. There's a couple of options for the implementation, so I thought I'd see what you prefer before submitting any PRs:

 1. Before processing an event, **rewrite REL_WHEEL events into SCROLLUP/SCROLLDOWN key events**. Whenever we emit a SCROLLUP/SCROLLDOWN, rewrite it back into a REL_WHEEL event
    - This is the smallest amount of code change, most of the code treats them as regular key presses.
    - This would end up rewriting any _actual_ SCROLLUP/SCROLLDOWN keyboard events into scroll events, even if you don't have any related mappings. That seems rude.
2. Introduce a **new struct with `key, value, event` fields**.
    - For key events the key and value would duplicate what's in the event, but for scroll events we could now represent `{ key: SCROLLUP, value: PRESS, event: RelAxis(REL_WHEEL, ...) }`. i.e. process mapping as if it were a SCROLLUP keypress event, but if it doesn't get remapped, emit the original event.
    - This has the lowest chance of breaking things, and the code seems clearer.
    - This type would need to be used in all the places we pass around `(key, value)` today, so it's not a small diff.
    - There's still some ambiguity in how we choose to emit a SCROLLUP which is produced by a remapping. We could probably leave it as a key, and hope it has the intended effect?
3. Introduce **an enum which is either a KeyEvent(key, value) or a ScrollEvent(direction)**
    - This is perhaps the most principled, but would require even more code changes.
    - This only makes sense if we want to treat scrollwheel events differently to SCROLLUP/SCROLLDOWN buttons. For that we'd have to invent new symbols like MOUSE_SCROLL_UP and MOUSE_SCROLL_DOWN, which means we couldn't represent them with a `Key` (which would require a lot of changes).

I'd like to implement option (2), but I'm happy to go with your preference.

k0kubun:
I've never played with the interface myself, so I can't give you a meaningful opinion. Can you go with an option that you think would be the simplest?

GitPerseus:
[...]

PR #181, opened 2022-09-28
Add support for scrollwheel events

timbertson:
As discussed in #180 , this is option (2). It's a slightly noisy diff because everything that as previously `(key, value)` has become a `MappableEvent`, to also carry around the original event.

The main logic change is that `main` always calls on_event, and then `event_handler` internally figures out whether it's a mappable event or whether to simply emit it.

### DIFF
--- a/src/event_handler.rs
+++ b/src/event_handler.rs
@@ -34,10 +34,11 @@ impl MappableEvent {
 
     pub fn from(raw: InputEvent) -> Option<Self> {
         match raw.kind() {
-            InputEventKind::Key(key) => {
-                Some(Self { key, value: raw.value(), raw })
-            },
-            
+            InputEventKind::Key(key) => Some(Self {
+                key,
+                value: raw.value(),
+                raw,
+            }),
             // Scrollwheel events are treated as if they were SCROLLUP/SCROLLDOWN keypresses,
             // but we emit the original event if it doesn't get remapped.
             InputEventKind::RelAxis(RelativeAxisType::REL_WHEEL) => {
@@ -46,11 +47,11 @@ impl MappableEvent {
                     -1 => Key::KEY_SCROLLDOWN,
                     other => {
                         debug!("Unknown scroll value: {}", other);
-                        return None
+                        return None;
                     }
                 };
                 Some(Self { key, value: PRESS, raw })
-            },
+            }
             _ => None,
         }
     }
@@ -205,7 +206,7 @@ impl EventHandler {
     fn dispatch_keys(
         &mut self,
         key_action: KeyAction,
-        event: MappableEvent
+        event: MappableEvent,
     ) -> Result<Vec<MappableEvent>, Box<dyn Error>> {
         let keys = match key_action {
             KeyAction::Key(modmap_key) => vec![MappableEvent::new(modmap_key, event.value)],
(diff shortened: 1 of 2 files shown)


k0kubun:
I'm out of home until this Saturday and don't have a Linux environment locally. Let me have a look after going back home 🙂

timbertson:
Of course, no worries!

k0kubun:
I have the same kind of impression on this PR as https://github.com/k0kubun/xremap/pull/187#issuecomment-1283459425. May I ask for a way to test something that this PR fixes and doesn't work on master?

timbertson:
Here's a sample config which this PR supports:

```yaml
keymap:
  - name: Tab shifting
    remap:
      LEFTALT-SCROLLDOWN: LEFTCTRL-PAGEDOWN
      LEFTALT-SCROLLUP: LEFTCTRL-PAGEUP
```


With this PR, I can hold down `alt` and use the mouse scrollwheel up/down to shift between tabs (in most applications). This is because it interprets a RELATIVE scroll event as SCROLLUP/SCROLLDOWN keys for the purposes of remapping (but if the key isn't remapped, it passes through the original scroll event).

src/event_handler.rs
```diff
@@ -19,6 +19,45 @@ use std::error::Error;
...
+            },
+            InputEventKind::RelAxis(RelativeAxisType::REL_WHEEL) => {
+                let key = match event.value() {
+                    1 => Key::KEY_SCROLLUP,
+                    -1 => Key::KEY_SCROLLDOWN,
+                    other => {
+                        debug!("Unknown scroll value: {}", other);
+                        return None
+                    }
+                };
```
GitPerseus, line 52: Bug: Depending on how fast you scroll, it is possible to create scroll events with values higher than 1 or lower than -1, which would cause scroll events to trigger the "other" branch of the match pattern.
I'd suggest replacing 1 and -1 with 1..=i32::MAX and i32::MIN..=-1 or something similar.

timbertson: That seems like a reasonable change, though I am surprised. In my testing I only ever saw 1/-1, other values only occurred in the HIGHRES event variants. I've only got a wired mouse though, are you using a trackpad?

GitPerseus: No, i'm using a wired mouse too (PERIDUO-406 wired mouse).
I do need to scroll somewhat fast to get values higher/lower than 1/-1, though.

k0kubun:

piit79:
Great work! I was just trying to remap my side scroll wheel that I used to use as volume up/down with `xbindkeys` (before I finally switched to Wayland...) and realized scroll events are not received/processed.

I'm a rust noob, but a long time dev and very interested in it, so I might take a look and try to help with the tests if you don't mind @timbertson ;)

k0kubun:
Feel free to open another PR, rebasing this branch :) As long as it has reasonable test cases in src/tests.rs, I'm open to merging the feature.

GitPerseus:
> Great work! I was just trying to remap my side scroll wheel
Does your side scroll wheel correspond to horizontal scrolling?
If so, then the event that you're interested in handling is REL_HWHEEL (and/or REL_HWHEEL_HI_RES),
which is different from the one currently handled in this branch (REL_WHEEL).

k0kubun:
Since https://github.com/k0kubun/xremap/pull/187 became ready first, I merged it and closed https://github.com/k0kubun/xremap/issues/180. Since the original issue is closed, I'm going to close this as well. Thank you for having a look at this!

timbertson:
Thanks, good work @GitPerseus

PR #187, opened 2022-10-06
Adds RELATIVE event catcher

GitPerseus:
The way i did this takes relative events and sends two fake key inputs to modmap.
(i added the fake key inputs to evdev's enum of key scancodes : https://github.com/GitPerseus/evdev/pull/1 ).

The first key input has a value of 1 and the second a value of 0, to simulate a instant press/unpress.

Note: My code isn't able to send relative event, only receive them.

It should also be possible to support more event types in about the same way.

### DIFF
--- a/src/action_dispatcher.rs
+++ b/src/action_dispatcher.rs
@@ -8,6 +8,7 @@ use nix::sys::signal;
 use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet};
 use std::process::{exit, Command, Stdio};
 
+use crate::event::RelativeEvent;
 use crate::{action::Action, event::KeyEvent};
 
 pub struct ActionDispatcher {
@@ -26,12 +27,42 @@ impl ActionDispatcher {
     }
 
     // Execute Actions created by EventHandler. This should be the only public method of ActionDispatcher.
-    pub fn on_action(&mut self, action: Action) -> anyhow::Result<()> {
-        match action {
-            Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
-            Action::InputEvent(event) => self.send_event(event)?,
-            Action::Command(command) => self.run_command(command),
-            Action::Delay(duration) => thread::sleep(duration),
+    pub fn on_action(&mut self, actions: Vec<Action>) -> anyhow::Result<()> {
+        //vector to collect all mouse movement events
+        let mut mouse_movement_actions: Vec<InputEvent> = Vec::new();
+        for action in actions {
+            match action {
+                Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
+                Action::RelativeEvent(relative_event) => self.on_relative_event(relative_event)?,
+                Action::MouseMovementEvent(mouse_movement_event) => {
+                    //collecting mouse movement events, to send them all in a single batch.
+                    mouse_movement_actions.push(InputEvent::new_now(
+                        EventType::RELATIVE,
+                        mouse_movement_event.code,
+                        mouse_movement_event.value,
+                    ));
+                }
+
+                Action::InputEvent(event) => self.send_event(event)?,
+                Action::Command(command) => self.run_command(command),
+                Action::Delay(duration) => thread::sleep(duration),
+            }
+        }
+
+        if mouse_movement_actions.len() != 0 {
+            //Sending all mouse movement events at once, unseparated by synchronization events.
+            self.send_mousemovement_event_batch(mouse_movement_actions)?;
+
+            //Mouse movement events need to be sent all at once because they would otherwise be separated by a synchronization event¹,
+            //which the OS handles differently from two unseparated mouse movement events.
+            //For example,
+            //a REL_X event², followed by a SYNCHRONIZATION event, followed by a REL_Y event³, followed by a SYNCHRONIZATION event,
+            //will move the mouse cursor by a different amount than
+            //a REL_X event followed by a REL_Y event followed by a SYNCHRONIZATION event.
+
+            //¹Because Xremap usually sends events one by one through evdev's "emit" function, which adds a synchronization event during each call.
+            //²Mouse movement along the X (horizontal) axis.
+            //³Mouse movement along the Y (vertical) axis.
         }
         Ok(())
     }
@@ -39,6 +70,16 @@ impl ActionDispatcher {
     fn on_key_event(&mut self, event: KeyEvent) -> std::io::Result<()> {
         let event = InputEvent::new_now(EventType::KEY, event.code(), event.value());
         self.send_event(event)
+    }
+
+    fn on_relative_event(&mut self, event: RelativeEvent) -> std::io::Result<()> {
+        let event = InputEvent::new_now(EventType::RELATIVE, event.code, event.value);
+        self.send_event(event)
+    }
+
+    //a function that takes mouse movement events to send in a single batch, unseparated by synchronization events.
+    pub fn send_mousemovement_event_batch(&mut self, eventbatch: Vec<InputEvent>) -> std::io::Result<()> {
+        self.device.emit(&eventbatch)
     }
 
     fn send_event(&mut self, event: InputEvent) -> std::io::Result<()> {
--- a/src/action_dispatcher.rs
+++ b/src/action_dispatcher.rs
@@ -27,42 +27,32 @@ impl ActionDispatcher {
     }
 
     // Execute Actions created by EventHandler. This should be the only public method of ActionDispatcher.
-    pub fn on_action(&mut self, actions: Vec<Action>) -> anyhow::Result<()> {
-        //vector to collect all mouse movement events
-        let mut mouse_movement_actions: Vec<InputEvent> = Vec::new();
-        for action in actions {
-            match action {
-                Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
-                Action::RelativeEvent(relative_event) => self.on_relative_event(relative_event)?,
-                Action::MouseMovementEvent(mouse_movement_event) => {
-                    //collecting mouse movement events, to send them all in a single batch.
-                    mouse_movement_actions.push(InputEvent::new_now(
-                        EventType::RELATIVE,
-                        mouse_movement_event.code,
-                        mouse_movement_event.value,
-                    ));
-                }
-
-                Action::InputEvent(event) => self.send_event(event)?,
-                Action::Command(command) => self.run_command(command),
-                Action::Delay(duration) => thread::sleep(duration),
+    pub fn on_action(&mut self, action: Action) -> anyhow::Result<()> {
+        match action {
+            Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
+            Action::RelativeEvent(relative_event) => self.on_relative_event(relative_event)?,
+            Action::MouseMovementEvent(_) => {
+                //This should not be reachable, since mouse movement events should collected
             }
-        }
+            Action::MouseMouvementEventCollection(mouse_movement_events) => {
+                //Sending all mouse movement events at once, unseparated by synchronization events.
+                self.send_mousemovement_event_batch(mouse_movement_events)?;
 
-        if mouse_movement_actions.len() != 0 {
-            //Sending all mouse movement events at once, unseparated by synchronization events.
-            self.send_mousemovement_event_batch(mouse_movement_actions)?;
+                //Mouse movement events need to be sent all at once because they would otherwise be separated by a synchronization event¹,
+                //which the OS handles differently from two unseparated mouse movement events.
+                //For example,
+                //a REL_X event², followed by a SYNCHRONIZATION event, followed by a REL_Y event³, followed by a SYNCHRONIZATION event,
+                //will move the mouse cursor by a different amount than
+                //a REL_X event followed by a REL_Y event followed by a SYNCHRONIZATION event.
 
-            //Mouse movement events need to be sent all at once because they would otherwise be separated by a synchronization event¹,
-            //which the OS handles differently from two unseparated mouse movement events.
-            //For example,
-            //a REL_X event², followed by a SYNCHRONIZATION event, followed by a REL_Y event³, followed by a SYNCHRONIZATION event,
-            //will move the mouse cursor by a different amount than
-            //a REL_X event followed by a REL_Y event followed by a SYNCHRONIZATION event.
+                //¹Because Xremap usually sends events one by one through evdev's "emit" function, which adds a synchronization event during each call.
+                //²Mouse movement along the X (horizontal) axis.
+                //³Mouse movement along the Y (vertical) axis.
+            }
 
-            //¹Because Xremap usually sends events one by one through evdev's "emit" function, which adds a synchronization event during each call.
-            //²Mouse movement along the X (horizontal) axis.
-            //³Mouse movement along the Y (vertical) axis.
+            Action::InputEvent(event) => self.send_event(event)?,
+            Action::Command(command) => self.run_command(command),
+            Action::Delay(duration) => thread::sleep(duration),
         }
         Ok(())
     }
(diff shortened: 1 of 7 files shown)


src/event_handler.rs
```diff
@@ -101,9 +102,19 @@ impl EventHandler {
                     continue;
                 }
             }
+            //checking if there's a "fake" key version of a relative event, (scancodes equal to and over 59974 are only "fake" custom events)
+            //and also if it's the same "key" and value as the one that came in.
+            if key.code() >= 59974 && (key.code(), value) == (event.code(), event.value()) {
```
k0kubun, line 107: Is there no way to avoid this magic number `59974` using something in the evdev crate? I think you could at least make it a const instead of writing that everywhere.

k0kubun:
You've explained this PR only with how it's implemented. Can you explain in what situation this is needed, e.g. remapping a mouse button while allowing a wheel to scroll a window? Is there a way to test the behavior you want to fix?

src/main.rs
```diff
@@ -211,7 +211,64 @@ fn handle_input_events(
...
+                            //since changing something by zero is kinda useless.
+                            //Just in case it can actually happen (and also because match arms need the same output type),
+                            //we'll just act like the value of the event was a positive.
+                            (event.code() * 2) + 59974
+                        }
+                    };
+
+                    //Sending a "fake" KEY event press to on_event.
+                    match handler
+                        .on_event(InputEvent::new(EventType::KEY, key, 1), config)
```
timbertson, line 251: nit: use `PRESS` / `RELEASE` instead of `1` / `0`

GitPerseus:
[Good point](https://github.com/k0kubun/xremap/pull/187#discussion_r998959676), i completely forgot abouts consts.
About not using it, my solution is based on using custom scancodes (to avoid conflating events),
so i need to offset relative events' values by some number to "start" at those custom scancode values.

It doesn't need to be 59974, though.
I just picked that number because then the highest custom scancode is 59999,
right below 60000 (the closest round number) and far away from 32768 (the previous power of two),
which makes it very unlikely that OS updates would add scancodes with conflicting values.




User case : 
Personally, i want to bind the mousewheel to keyboard keys because there are quite a few games that have "next item"/"previous item" buttons,
but no way to remap those to the mousewheel, when it feels much more logical (and natural) to me.

I might have a use case for mouse movement remapping (a game).
For other events, i don't have a use case (i don't even know what half of them are supposed to do or what would ever emit them),
it's just that generalizing my solution for scroll events to implement them was very easy.




Test case :
With the following config file (while capturing the mouse) :

```
modmap:
  - name: Global
    remap:
      BTN_XUPSCROLL: KEY_E
      BTN_XDOWNSCROLL: KEY_A
      BTN_XRIGHTCURSOR: KEY_P

```
It's possible to test that scrolling translates to the corresponding keys.

k0kubun:

GitPerseus:
I have completed the rebase.
Everything works pretty much the same way it did before,
i just had to add a few enums variants/change some types here and there.

I also added [one unit test](https://github.com/GitPerseus/xremap/blob/feature/adds_relative_event_catcher/src/tests.rs#L74).
Should i add more of the same, or did you have something else in mind?


I also fixed a bug.
Here's two scenarios :
-Mouse movement events separated by a SYNCHRONIZATION event;
-The same mouse movement events unseparated by a SYNCHRONIZATION event.
The OS treats them differently.

I have made two tests¹ that demonstrate this difference,
[test_cursor_behavior_1 and test_cursor_behavior_2](https://github.com/GitPerseus/xremap/blob/feature/adds_relative_event_catcher/src/tests.rs#L89-L197).

¹These tests are just meant to demonstrate the difference,
they don't do anything to check if the bug currently exists or not.

Xremap usually emits events one by one,
which separates events with synchronization events,
regardless of pre-existing synchronization events.

This means that Xremap separates mouse movement events that would normally be consecutive,
which causes a captured mouse to act differently,
which is a bug.

To prevent that,
i made on_action take a vector of all actions from the latest call to fetch_events,
to be able to emit all mouse movement events at once,
which avoids separating them with synchronization events.

src/action_dispatcher.rs
```diff
@@ -40,6 +71,16 @@ impl ActionDispatcher {
         self.send_event(event)
     }
 
+    fn on_relative_event(&mut self, event: RelativeEvent) -> std::io::Result<()> {
+        let event = InputEvent::new_now(EventType::RELATIVE, event.code, event.value);
+        self.send_event(event)
+    }
+
+    //a function that takes mouse movement events to send in a single batch, unseparated by synchronization events.
+    pub fn send_mousemovement_event_batch(&mut self, eventbatch: Vec<InputEvent>) -> std::io::Result<()> {
```
k0kubun, line 80: Can you please not make this `pub`? L27 specifically asks you: "This should be the only public method of ActionDispatcher."

src/action_dispatcher.rs
```diff
@@ -25,12 +26,42 @@ impl ActionDispatcher {
     }
 
     // Execute Actions created by EventHandler. This should be the only public method of ActionDispatcher.
-    pub fn on_action(&mut self, action: Action) -> anyhow::Result<()> {
-        match action {
-            Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
-            Action::InputEvent(event) => self.send_event(event)?,
-            Action::Command(command) => self.run_command(command),
-            Action::Delay(duration) => thread::sleep(duration),
+    pub fn on_action(&mut self, actions: Vec<Action>) -> anyhow::Result<()> {
```
k0kubun, line 29: You should be able to emit a vector of `MouseMovementEvent` as a type of Action and handle it here instead of changing the argument type to `actions: Vec<Action>`. The idea of `EventHandler` / `ActionDispatcher` is to isolate the portion that has side effects and is harder to test (`ActionDispatcher`) from the (basically side-effect-free) pure function (`EventHandler`). Every complicated event transformation should be done in `EventHandler` so it can be tested, and `ActionDispatcher` should be dumb enough to obviate tests.

src/config/key.rs
```diff
@@ -47,6 +47,64 @@ pub fn parse_key(input: &str) -> Result<Key, Box<dyn Error>> {
         "SUPER_L" => Key::KEY_LEFTMETA,
         "WIN_R" => Key::KEY_RIGHTMETA,
         "WIN_L" => Key::KEY_LEFTMETA,
+
+        //Entirely custom scancodes
```
k0kubun, line 51: Where did they come from? I'm not sure if I can maintain this portion. Is it possible to send this patch to evdev crate instead?

GitPerseus: Oh right, i only ever mentioned that change [here](https://github.com/k0kubun/xremap/issues/180#issuecomment-1288149131).

It's a way for Xremap to recognize names like "BTN_XRIGHTCURSOR" in config files as valid,
which is much simpler than modifying evdev,
and is also easier to modify.

In other words,
my code no longer uses a change in evdev's scancode enum,
but instead gives custom aliases to scancodes,
just like the code above said aliases.

An example for clarity :
`"WIN_L" => Key::KEY_LEFTMETA,`
is basically the same thing as
`"WIN_L" => evdev::Key(125),`
The first just uses the name KEY_LEFTMETA instead of the corresponding scancode.

Here's one of my custom aliases :
`"BTN_XRIGHTCURSOR" => evdev::Key(59974),`
There is no name corresponding to that scancode,
so my code has to use the scancode instead.


My code takes relative events,
"transforms" them into key events with those scancodes,
and then runs those key events through modmap and keymap.

To be able to map those key events to other events,
users need names to put in the config file,
which is where these aliases are used.

The reason the scancodes i use have no corresponding name is because no device emits them,
which is perfect,
since otherwise modmap/keymap wouldn't be able to tell the difference between those events and my disguised events.


I'll change my very undescriptive and misleading comment to better reflect what those aliases are and why they're used.



I'm not sure i see any possible problems maintaining these aliases,
since maintaining them and other aliases should be no different.

It should be possible to send the equivalent of this as a pull request to the evdev crate instead,
but they might not accept it,
and since i don't see why these aliases would be difficult to maintain,
I'm not sure that would be very useful.

k0kubun: > Here's one of my custom aliases

Why do you need a custom alias?

> The reason the scancodes i use have no corresponding name is because no device emits them, which is perfect, since otherwise modmap/keymap wouldn't be able to tell the difference between those events and my disguised events.

Why do you need to disguise events?

k0kubun:
Please rebase your branch again and address other feedback as well. Thank you :pray:

src/config/key.rs
k0kubun: Perhaps you can add a config syntax to define those custom codes. I don't think xremap should be responsible for maintaining your original scancodes.

GitPerseus: (I'm assuming that you're still talking about custom aliases?)

>Perhaps you can add a config syntax to define those custom codes. 

I'm not sure i understand what you're proposing.

How would config syntaxes allow giving names to scancodes,
names usable in config files?
What's the difference between my custom aliases and the ones above?
As in, why is it better for config syntax to replace my custom aliases,
but not other custom aliases?

I can't think of a way to use config syntax that wouldn't lose the clarity these custom aliases can have.
For instance, which config file do you think is clearest :

The one using some config syntax to differentiate between positive and negative relative events?
```
modmap:
  - name: config syntax
    remap:
      positive: REL_X: KEY_H
      negative: REL_X: KEY_J
      
      positive: REL_WHEEL: KEY_E
      negative: REL_WHEEL: KEY_T
      
      positive: REL_Y: KEY_K
      negative: REL_Y: KEY_L
```

or the one using custom aliases to do the same thing?
```
modmap:
  - name: custom aliases
    remap:
      BTN_XUPCURSOR: KEY_H
      BTN_XDOWNCURSOR: KEY_J
      
      BTN_XDOWNSCROLL: KEY_E 
      BTN_XUPSCROLL: KEY_T
      
      BTN_XLEFTCURSOR: KEY_K
      BTN_XRIGHTCURSOR: KEY_L
```



>I don't think xremap should be responsible for maintaining your original scancodes.

You do maintain other custom aliases,
what would be different about mine?
If it's that you think the numbers¹ seem arbitrary and could cause confusion,
I could use the [offsetter const](https://github.com/GitPerseus/xremap/blob/feature/adds_relative_event_catcher/src/event_handler.rs#L139) to define them,
and even create a test to make sure the const isn't changed to a problematic number.


¹(the numbers used instead of names like in other custom aliases.)

GitPerseus: (I'm assuming that you're still talking about custom aliases?)

>Perhaps you can add a config syntax to define those custom codes. 

I'm not sure i understand what you're proposing.

How would config syntaxes allow giving names to scancodes,
names usable in config files?
What's the difference between my custom aliases and the ones above?
As in, why is it better for config syntax to replace my custom aliases,
but not other custom aliases?

I can't think of a way to use config syntax that wouldn't lose the clarity these custom aliases can have.
For instance, which config file do you think is clearest :

The one using some config syntax to differentiate between positive and negative relative events?
```
modmap:
  - name: config syntax
    remap:
      positive: REL_X: KEY_H
      negative: REL_X: KEY_J
      
      positive: REL_WHEEL: KEY_E
      negative: REL_WHEEL: KEY_T
      
      positive: REL_Y: KEY_K
      negative: REL_Y: KEY_L
```

or the one using custom aliases to do the same thing?
```
modmap:
  - name: custom aliases
    remap:
      BTN_XUPCURSOR: KEY_H
      BTN_XDOWNCURSOR: KEY_J
      
      BTN_XDOWNSCROLL: KEY_E 
      BTN_XUPSCROLL: KEY_T
      
      BTN_XLEFTCURSOR: KEY_K
      BTN_XRIGHTCURSOR: KEY_L
```



>I don't think xremap should be responsible for maintaining your original scancodes.

You do maintain other custom aliases,
what would be different about mine?
If it's that you think the numbers¹ seem arbitrary and could cause confusion,
I could use the [offsetter const](https://github.com/GitPerseus/xremap/blob/feature/adds_relative_event_catcher/src/event_handler.rs#L139) to define them,
and even create a test to make sure the const isn't changed to a problematic number.


¹(the numbers used instead of names like in other custom aliases.)

k0kubun: > What's the difference between my custom aliases and the ones above?
> As in, why is it better for config syntax to replace my custom aliases, but not other custom aliases?

I think the point is, I believe existing custom aliases are something many people use, and I still don't understand why you would want custom aliases for your case.

Are they something not defined by evdev at all? Why can't we define them in evdev? The worst case scenario is that we define custom names first, the upstream evdev defines them later, and we end up needing to maintain both inconsistent names for backward compatibility. Are you sure it's not gonna happen?

k0kubun: Also, can you consider addressing other comments I left? If the code is in a mergeable state, I might just merge it, but somehow you keep doing discussions. Reading code that is consistent with existing code and tests that verify what you want to do are sometimes more expressive than discussions.

GitPerseus:
>Also, can you consider addressing other comments I left first?

Sorry about that.
I left thumbs up emojis on them to indicate that i agreed with those comments,
forgetting that emojis don't send notifications.
(I have made changes as you suggested.)


>The worst case scenario is that we define custom names first, the upstream evdev defines them later, and we end up needing to maintain both inconsistent names for backward compatibility. Are you sure it's not gonna happen?

Pretty sure, yes.
Currently, the highest number for a scancode is 743,
with a total of ~547 defined scancodes (some numbers are skipped, not sure why).
The numbers i've chosen to define my disguised events with starts at 59974,
which, at the current rate,
would only be reached by upstream evdev in over 80 times its lifetime.
Of course, "the current rate" is inaccurate.
In reality, i think it would actually take even longer, if it ever happened.
And even then, because i'm defining the scancodes for those disguised events with a const,
you'd just need to change the const to a different offset and the problem would be fixed.


>Code that is consistent with existing implementation and tests that verify what you want to do are sometimes more expressive than discussions.

Hopefully the code within this commit will answer your remaining questions.

src/action_dispatcher.rs
```diff
@@ -29,6 +30,26 @@ impl ActionDispatcher {
     pub fn on_action(&mut self, action: Action) -> anyhow::Result<()> {
         match action {
             Action::KeyEvent(key_event) => self.on_key_event(key_event)?,
+            Action::RelativeEvent(relative_event) => self.on_relative_event(relative_event)?,
+            Action::MouseMovementEvent(_) => {
+                //This should not be reachable, since mouse movement events should collected
```
k0kubun, line 35: Can you call `unreachable!()` here then?

src/main.rs
```diff
@@ -224,32 +224,47 @@ fn handle_input_events(
...
+            //A vector used to collect all mouse movement events
+            let mut mousemovements: Vec<RelativeEvent> = Vec::new();
+
             for event in events {
                 if let Some(event) = Event::new(event) {
-                    handle_event(handler, dispatcher, config, event)?;
+                    //Running the event through modmap and keymap
+                    let handled_events = handle_event(handler, config, event)?;
+                    //and checking if the resulting events are mouse movements
+                    for action in handled_events {
```
k0kubun, line 235: I think you needed to aggregate mouse movement events here because the event handler accepts events one by one. However, as I explained before, the point of the current code structure is to make it testable by just examining the input and the output of the event handler. If you implement this logic outside it, then it's kind of nullified.

WDYT about changing `handle_event` / `on_event` to `handle_events` / `on_events`? Then I think you can handle this inside the event handler.

src/main.rs
```diff
@@ -224,32 +224,47 @@ fn handle_input_events(
         Err((Some(ENODEV), _)) => Ok(false),
         Err((_, error)) => Err(error).context("Error fetching input events"),
         Ok(events) => {
+            //A vector used to collect all mouse movement events
```
k0kubun, line 227: nitpick: Since existing code uses the comment style `// comment`, could you consider putting a space (` `) after `//`? It's nice to have consistency across the codebase.

k0kubun:
Thanks for working on this again! I left comments that will hopefully be the last feedback.

GitPerseus:
I have formatted my comments and changed `handle_event`/`on_event` to `handle_events`/`on_events`.
Because of the latter,
i didn't need `MouseMovementEvent` anymore and got rid of the unreachable branch,
and was able to add a test making sure that mouse movements are being aggregated.

src/main.rs
```diff
@@ -57,7 +56,7 @@ struct Opts {
         verbatim_doc_comment,
         hide_possible_values = true,
         // Separating the help like this is necessary due to
-        // https://github.com/clap-rs/clap/issues/3312
+        // https:// github.com/clap-rs/clap/issues/3312
```
k0kubun, line 59:
```suggestion
        // https://github.com/clap-rs/clap/issues/3312
```

k0kubun (approved):
Thank you for addressing my feedback!

Issue #193, opened 2022-10-18
Absolute input support for tablets


ToxicMushroom:
Hi, I use xorg with i3.
I wanted to try xremap to map my tablet buttons to specific shortcuts for xournalpp.
After configuring my button remaps I noticed that my pen doesn't work anymore xd, after looking into the code I noticed that the virtualdevice doesn't register any absolute axes.
When I added these axes locally xorg still ignored the uinput events.

I doubt you have a tablet to test this so I'm open for providing more information, if this is solvable at all because I'm a bit stumped

ToxicMushroom:
**evtest /dev/input/event262**
```
Input driver version is 1.0.1
Input device ID: bus 0x3 vendor 0x256c product 0x6d version 0x111
Input device name: "GAOMON Gaomon Tablet"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
    Event code 320 (BTN_TOOL_PEN)
    Event code 330 (BTN_TOUCH)
    Event code 331 (BTN_STYLUS)
    Event code 332 (BTN_STYLUS2)
  Event type 3 (EV_ABS)
    Event code 0 (ABS_X)
      Value  28526
      Min        0
      Max    33020
      Resolution     200
    Event code 1 (ABS_Y)
      Value   2112
      Min        0
      Max    20320
      Resolution     200
    Event code 24 (ABS_PRESSURE)
      Value      0
      Min        0
      Max     8191
    Event code 26 (ABS_TILT_X)
      Value      0
      Min      -60
      Max       60
      Resolution      57
    Event code 27 (ABS_TILT_Y)
      Value      0
      Min      -60
      Max       60
      Resolution      57
  Event type 4 (EV_MSC)
    Event code 4 (MSC_SCAN)
Properties:
  Property type 0 (INPUT_PROP_POINTER)
Testing ... (interrupt to exit)
***********************************************
  This device is grabbed by another process.
  No events are available to evtest while the
  other grab is active.
  In most cases, this is caused by an X driver,
  try VT-switching and re-run evtest again.
  Run the following command to see processes with
  an open fd on this device
 "fuser -v /dev/input/event262"
***********************************************
```

**evtest /dev/input/event260**
```Input driver version is 1.0.1
Input device ID: bus 0x3 vendor 0x1234 product 0x5678 version 0x111
Input device name: "xremap pid=72851"
Supported events:
  Event type 0 (EV_SYN)
  Event type 1 (EV_KEY)
[...]

PR #194, opened 2022-10-19
feat: support absolute input devices like tablets

ToxicMushroom:
This aims to solve #193 
Feel free to comment on style issues or other things :)

I tested the pr locally and seems to work very well for my usecase

### DIFF
--- a/example/config.yml
+++ b/example/config.yml
@@ -179,7 +179,6 @@ keymap:
       Super-d: M-d
       Super-i: M-i
       Super-w: M-w
-      Super-w: M-w
       Super-t: M-t
       Super-c: M-c
       Super-v: M-v
@@ -213,35 +212,3 @@ keymap:
       Super-7: C-7
       Super-8: C-8
       Super-9: C-9
-
-# Absolute config section for tablet devices or touchscreens.
-# Missing fields default to 0,
-# so you don't have to set them if the desired value is 0
-absolute:
-
-  # Below is an example config for a GAOMON S620.
-  # x and y decide what area of your tablet is mapped
-  x:
-    value: 0
-    min: 0
-    max: 33020
-    fuzz: 0
-    flat: 0
-    resolution: 200
-  y:
-    min: 0
-    max: 20320
-    resolution: 200
-
-  # pressure, x_tilt and y_tilt are pen properties.
-  pressure:
-    min: 0
-    max: 8191
-  tilt_x:
-    min: -60
-    max: 60
-    resolution: 57
-  tilt_y:
-    min: -60
-    max: 60
-    resolution: 57
--- a/example/config.yml
+++ b/example/config.yml
@@ -213,3 +213,35 @@ keymap:
       Super-7: C-7
       Super-8: C-8
       Super-9: C-9
+
+# Absolute config section for tablet devices or touchscreens.
+# Missing fields default to 0,
+# so you don't have to set them if the desired value is 0
+absolute:
+
+  # Below is an example config for a GAOMON S620.
+  # x and y decide what area of your tablet is mapped
+  x:
+    value: 0
+    min: 0
+    max: 33020
+    fuzz: 0
+    flat: 0
+    resolution: 200
+  y:
+    min: 0
+    max: 20320
+    resolution: 200
+
+  # pressure, x_tilt and y_tilt are pen properties.
+  pressure:
+    min: 0
+    max: 8191
+  tilt_x:
+    min: -60
+    max: 60
+    resolution: 57
+  tilt_y:
+    min: -60
+    max: 60
+    resolution: 57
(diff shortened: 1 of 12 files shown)


src/main.rs
```diff
@@ -208,15 +208,18 @@ fn handle_input_events(
...
         Ok(events) => {
+            let mut vec: Vec<InputEvent> = Vec::new();
             for event in events {
                 if event.event_type() == EventType::KEY {
                     handler
                         .on_event(event, config)
                         .map_err(|e| anyhow!("Failed handling {event:?}:\n  {e:?}"))?;
                 } else {
-                    handler.send_event(event)?;
+                    vec.push(event);
```
k0kubun, line 218: This means that, if you interleave key and non-key events, they could be reordered. Is that intentional? I feel like you should push multiple events at once only for the ranges that don't include the other type of events.

src/device.rs
```diff
@@ -55,10 +67,24 @@ pub fn output_device() -> Result<VirtualDevice, Box<dyn Error>> {
     relative_axes.insert(RelativeAxisType::REL_WHEEL);
     relative_axes.insert(RelativeAxisType::REL_MISC);
 
+    let info = AbsInfo::new(0, 0, 25000, 0, 0, 200);
+    let tilt_info = AbsInfo::new(0, 0, 60, 0, 0, 57);
+    let press_info = AbsInfo::new(0, 0, 8191, 0, 0, 0);
```
k0kubun, line 72: I'm not so sure about how I could maintain those non-zero numbers (`25000`, `200`, `60`, `57`, `8191`). Can you leave a comment that has a link to some reference or use a library-provided constant like the following variables?

src/main.rs
ToxicMushroom: True I didn't think of the reordering.
Shkuld I send them in bulk only when absolute input is detected?

src/device.rs
ToxicMushroom: Sorry for not testing this before, it appears these values decide what area of the tablet is usable an how accurate the movement is. The current values were just a copy paste from my tablet's file descriptor but it would be desirable to make them configurable in the config.

ToxicMushroom:
Eh I think it's unsupported for a virtual device to do both absolute, mouse and keyboard inputs. 
I think linux (at least my distro) doesn't know what drivers to use for a device with both keys, relative and absolute input.
When I comment out the with_absolute_axis methods in device.rs then keyboard presses are received by xorg ect, otherwise they aren't.

I'll do more research tomorrow if I find time, maybe it's possible to have multiple virtual devices but I am unsure.

example/config.yml
```diff
@@ -213,3 +213,35 @@ keymap:
       Super-7: C-7
       Super-8: C-8
       Super-9: C-9
+
+# Absolute config section for tablet devices or touchscreens.
+# Missing fields default to 0,
+# so you don't have to set them if the desired value is 0
+absolute:
```
k0kubun, line 220: I want this example to be a separate file because keyboards are still the primary use case of xremap. You may put something like `tablet.yml`.

Also, please consider updating `src/config/tests.rs`.

ToxicMushroom:
Will stall until https://github.com/emberian/evdev/pull/106 and https://github.com/emberian/evdev/pull/105 are handled

ToxicMushroom:
I added a deviceType enum and targets to modmap and keymap
This is to have events come from non tablets and potentially send them to the main virtual device or tablet virtual device.
or from other input devices to tablet ect. 
I don't know if the implementation I did does a good job at accomplishing this.

k0kubun:

ToxicMushroom:
Sure, but we still need to wait for evdev to publish to crates.io I think ?

k0kubun:
Ah yes, I just got confused about that. I made this a draft PR. Please mark it ready for reviews when the crate is ready too :+1:

ToxicMushroom:
should I resolve the conflicts before your review it ?

wvffle:
Hi! I know the PR is quite old but I'd like to share some input if I may.

Analog events are spawned also by other devices than tablets. Maybe it's a good idea to add a generic support for absolute events?

I'd want to remap some keys on my Steam Deck, it also has some absolute inputs like:
- `ABS_HAT0X` , `ABS_HAT0Y` for left trackpad
- `ABS_HAT1X`, `ABS_HAT1Y for right trackpad
- `ABS_HAT2Y` for left trigger
- `ABS_HAT2X` for right trigger
- `ABS_X`, `ABS_Y`, `ABS_RX`, `ABS_RY` - idk, perhaps gyro. The log is flooded with these events.
- The joysticks do not spawn `EV_ABS` events for some reason.

k0kubun:
Feel free to open another PR. I don't know how to implement them, but it seems like you do.

PR #196, opened 2022-10-22
fix: libinput dwt support

### LINKED ISSUE #152: Support dwt for libinput
This is not really a bug of xremap I think, but a configuration problem.

# Problem

With xremap activated, `dwt` libinputs "disable-while-typing" is not working.

# Reproduce

Run `xremap` on a laptop and enable `dwt` (e.g. for sway, you do that by adding `input * dwt enabled` to your config).

While typing, try to move your mouse with the touchpad, if it was working correctly your touchpad should not be controlling your mouse (this should also be the case without `xremap` running).

On my device, the touchpad disabling while typing did not work when `xremap` is running, but did work when I killed `xremap`.

# Proposed solution
Following https://gitlab.freedesktop.org/libinput/libinput/-/issues/524 the solution would be to convince `libinput` that the `xremap` keyboard is an internal keyboard if the source is.

Maybe this requires manual intervention to just put something similar to Depau in the `/etc/libinput/local-overrides.quirks`:
```ini
[xremap]
MatchUdevType=keyboard
MatchBus=usb //or whatever we need
MatchName= // whatever xremap uses
AttrKeyboardIntegration=internal
```

ModProg:
Fixes #152

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -4,7 +4,7 @@ extern crate nix;
 use anyhow::bail;
 use derive_where::derive_where;
 use evdev::uinput::{VirtualDevice, VirtualDeviceBuilder};
-use evdev::{AttributeSet, Device, FetchEventsSynced, Key, RelativeAxisType};
+use evdev::{AttributeSet, BusType, Device, FetchEventsSynced, InputId, Key, RelativeAxisType};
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
Thank you!

Issue #197, opened 2022-10-22
trigger issue 0.7.7.1


cipherlogs:
related to #154 

for example 

shift + win + k will trigger Win - k (which is up in my config)
ctrl + win + k will trigger win - k (which is up in my config)

```
keymap:
  - name: Global
    remap:
      WIN-k: up
```

k0kubun:
Can you write both the current behavior and the expected behavior? I don't understand which one you wrote.

cipherlogs:
the expected behaviour is to press shift + win + k instead of pressing win + k

k0kubun:
Sorry, that was a bad way to ask it. Can you provide the following three bullets? e.g.

* Input keys: Shift + Win + K
* Current behavior: Win + K
* Expected behavior: Shift + Win + K

If this is exactly what you meant, please copy-paste it as is.

cipherlogs:
Yes that's the exact behavior

* Input keys: Shift + Win + K
* Current behavior: Win + K
* Expected behavior: Shift + Win + K

k0kubun:
I can't reproduce your "Current behavior". This is my output of `RUST_LOG=debug xremap config.yml`:

```
[2022-11-03T03:52:24Z DEBUG xremap::event_handler] => 0: KEY_ENTER
[2022-11-03T03:52:24Z DEBUG xremap::action_dispatcher] 0: KEY_ENTER
[2022-11-03T03:52:26Z DEBUG xremap::event_handler] => 1: KEY_RIGHTMETA
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 1: KEY_RIGHTMETA
[2022-11-03T03:52:26Z DEBUG xremap::event_handler] => 1: KEY_LEFTSHIFT
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 1: KEY_LEFTSHIFT
[2022-11-03T03:52:26Z DEBUG xremap::event_handler] => 1: KEY_K
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 0: KEY_RIGHTMETA
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 1: KEY_UP
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 0: KEY_UP
[2022-11-03T03:52:26Z DEBUG xremap::action_dispatcher] 1: KEY_RIGHTMETA
[...]

Issue #102, opened 2022-05-24
Automaticaly adding modifiers in keymap


danyspin97:
I am replacing `rightalt + n` with `left` key and a couple more. However, it requires a new remap for every added modifier combination.

```
keymap:
  - name: Arrows
    remap:
      alt_r-n: left
      alt_r-e: down
      alt_r-i: up
      alt_r-o: right
      super-alt_r-n: super-left
      super-alt_r-e: super-down
      super-alt_r-i: super-up
      super-alt_r-o: super-right
      super-shift_l-alt_r-n: super-shift_l-left
      super-shift_l-alt_r-e: super-shift_l-down
      super-shift_l-alt_r-i: super-shift_l-up
      super-shift_l-alt_r-o: super-shift_l-right
```

Is there an way to just add the first four remaps and make them work with any other modifier automatically? I have tried with `modmap` but I couldn't make it work.

k0kubun:
Yeah, I think `alt_r-n: left` alone should remap `whatever-alt_r-n` to `whatever-left`. It's currently not working that way due to a premature optimization in the implementation.

I'll work on this when I can allocate time to work on this. Meanwhile, please use the current workaround and also consider generating it with a template as needed.

k0kubun:
Update: As I'm currently focusing on a big change for another OSS, I'm planning to work on this issue on Aug 13th. If I receive a pull request before that, I can take a look at that too.

k0kubun:
Released https://github.com/k0kubun/xremap/pull/145 in v0.6.0. Enjoy.

nullbyto:
For report: i tried it with release v0.7.0 and downloaded v0.6.0, it's not working for me. I'm not sure if I have to do something extra.

k0kubun:
In addition to the version, could you report the following information?

* Platform: X11, GNOME Wayland, or Sway
* Your config (preferably minimum one)
* Output of `RUST_LOG=debug xremap`

For example, using xremap v0.7.0 on X11, my config to test this behavior is:

```yml
keymap:
  - remap:
      alt_r-n: left
```

and the `RUST_LOG=debug xremap` is:

```
[2022-08-14T22:23:35Z DEBUG xremap::event_handler] => 1: KEY_LEFTCTRL
[2022-08-14T22:23:35Z DEBUG xremap::event_handler] 1: KEY_LEFTCTRL
[2022-08-14T22:23:35Z DEBUG xremap::event_handler] => 1: KEY_RIGHTALT
[...]

PR #209, opened 2022-12-05
Introduce keymap level `exact_match` option

laech:
Related to https://github.com/k0kubun/xremap/issues/197 and https://github.com/k0kubun/xremap/issues/102

Example:

```yaml
  keymap:
    - exact_match: true
      remap:
        M-f: C-right
```

Given the above, and <kbd>M-Shift-f</kbd> is pressed:

- If `exact_match` is false or unset, the existing behaviour will be used,
  which will translate <kbd>M-Shift-f</kbd> to <kbd>C-Shift-right</kbd>.

- If `exact_match` is true, <kbd>M-Shift-f</kbd> will be sent as is, i.e. not matched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ is supported only in `modmap` since `keymap` handles modifier keys differently.
 ```yml
 modmap:
   - name: Name # Optional
+    exact_match: false # Optional, defaults to false
     remap: # Required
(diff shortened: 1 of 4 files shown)


k0kubun (approved):
Thanks!

PR #212, opened 2022-12-17
Reconnect to X11 when existing connection stops working

laech:
I have been having some occasionally issues where xremap would stop working due to the existing connection to X11 breaks.

This issue can be reproduced reliably by running xremap as a global systemd service, then logout and log back in, to see xremap still runs but no longer remaps any keys or would also eat some of the keys.

So this PR will attempt to reconnect to X11 when we receive an error when using the X11 connection.

### DIFF
--- a/src/client/x11_client.rs
+++ b/src/client/x11_client.rs
@@ -1,7 +1,11 @@
 use crate::client::Client;
+use anyhow::bail;
 use std::env;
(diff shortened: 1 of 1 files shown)


k0kubun (approved):
:ok_hand:

laech:
Thanks for the quick response!

PR #214, opened 2022-12-20
Ensure the launched commands are not terminated by the stop of xremap.

### LINKED ISSUE #213: Child process killed when xremap restarts
Heya,

First off, thanks for your great work my good sir! This works really well for me, it's really good to be able to configure all the key mappings at will.

However, I find a minor inconvenience with the launch directive in the config file, the launched processes are a direct child process of xremap, if we stop xremap for whatever reason(debug/update config for example), the child process dies. Which might not be ideal.

Are you open to use a crate like [fork](https://crates.io/crates/fork) to daemonize the subprocess? Thanks!

xiaket:
Heya it's me again. :D

Creating this PR to fix #213.

Hope you won't mind some notes on the implementation: we cannot use the `daemon` function provided by [fork](https://crates.io/crates/fork), because the parent [will be terminated](https://github.com/immortal/fork/blob/d82131f1668ad292e58bf6de62e2e51fa1efc0f0/src/lib.rs#L178). So instead, I've use the more primitive `fork` API to do the double fork. In the first fork, the parent(grandfather) will do nothing and continue, the child process(father) will fork again. And this time, the parent(father) will simply exit 0, and the child process(child) will spawn the command then exit 0. We forked twice and exited twice, leaving the original grandfather process untouched.

I'ved tested this by doing a release build. Before the change, if I started xremap and launched a browser, then kill xremap, browser will be terminated. After this change, the launched browser process will not be killed.

One last note on the setup, it looks like even with this change in place, the default systemd setup will terminate everything in the control group, which means the subprocess in the new session/process group will be killed as well. So, to stop systemd from killing every thing, I would need to set `KillMode` to `process` in my systemd service file.


k0kubun (approved):
Thanks!

> So, to stop systemd from killing every thing, I would need to set KillMode to process in my systemd service file.

Good to know :)

PR #216, opened 2022-12-22
Feature hyprland

chenzaichun:
Init Hyprland support.


k0kubun (approved):
Wow, that was quick. Thank you!

PR #220, opened 2022-12-28
fix: application None check when failed get the application name

chenzaichun:


src/client/hypr_client.rs
```diff
@@ -19,7 +19,9 @@ impl Client for HyprlandClient {
             let s = serde_json::to_string(&win).ok()?;
             let v: Value = serde_json::from_str(&s).ok()?;
             let app = v["class"].as_str();
-            return Some(String::from(app.unwrap()));
+            if app.is_some() {
+                return Some(String::from(app.unwrap()));
+            }
```
k0kubun, line 24:
```suggestion
            if let Some(app) = app {
                return Some(String::from(app));
            }
```

k0kubun (approved):

Issue #250, opened 2023-02-27
kde wayland


N4tus:
Is there support to use xremap on kde with wayland?

k0kubun:
First off, it should already work except for `application` filter. If you're not interested in using `application` filtering, you should have no problem with it.

If you do have interest in using `application` filter, we'd have to find a protocol to make that happen. Wayland is unfortunately still too immature to have that capability at that layer. So we rely on another layer on top of it. We currently target individual composers, but it's planned to switch Sway and Hypr to use wlroots protocol instead. Many Wayland composers use wlroots protocol and it seems like `application` filter can be implemented with that protocol.

However, some composers like GNOME don't use wlroots. I'm not sure if KDE uses wlroots, but I suspect it might not. Do you know of any protocol used by KDE with Wayland that could power `application` filter?

N4tus:
Yes KDE does not use wlroots. There are interfaces that allow applications to communicate with kwin, but I do not know what you can do with them and how to interact with it.

N4tus:
After digging for some protocols [`org_kde_plasma_window_management`](https://github.com/KDE/plasma-wayland-protocols/blob/master/src/protocols/plasma-window-management.xml) seems like a promising candidate, especially the `org_kde_plasma_activation_feedback` interface.

k0kubun:
Nice. Would you be interested in working on a PR to support it?

RedBearAK:
I subscribed to this issue even though I don't really have anything to contribute yet. I'm also interested in finding out if there is a way to get the window class and name/title in KDE while using Wayland. Information online regarding this is limited. 

I've been able to successfully connect to two different GNOME extensions via D-Bus from a Python app to get this information in Wayland. If it's being presented by some D-Bus interface in Plasma, the same thing should be possible with a Plasma script. 

These look interesting, but I don't know if this protocol just provides event signals or would actually be able to provide the window title and application class/name/id. 

```xml

    <event name="title_changed">
      <description summary="window title has been changed">
        This event will be sent as soon as the window title is changed.
      </description>
      <arg name="title" type="string" summary="window title"/>
    </event>

    <event name="app_id_changed">
      <description summary="application identifier has been changed">
        This event will be sent as soon as the application
        identifier is changed.
      </description>
      <arg name="app_id" type="string"/>
    </event>

```

And this part, although I'm bothered by the "only one client can bind to this interface at a time" part. 

```xml
  <interface name="org_kde_plasma_window_management" version="16">
[...]

PR #264, opened 2023-03-20
Kde wayland

N4tus:
A KDE-Plasma client for wayland (should also run on x11 but I have not tested that).  
The only strange thing is, that the system-log report a `Type Error: Type Error` each time the script react to a active window change. But it only does a DBusCall and this is successfully executed, so I don't know where the error should originate.

The PR for issue #250


k0kubun (approved):
:+1:

PR #269, opened 2023-03-28
handle multiple config files

### LINKED ISSUE #202: [Feature request] Import files in config
It would be nice if we could import files in a config file to reduce duplication. Here's an idea of the syntax:
```yml
import:
  - name: application-specific
    file1.yml # the extension is optional
    file2     # imports file2.yml

modmap:
  ...

keymap:
  ...
```

This would be useful for people who have keyboards with different layouts. For example, I have a MacBook Air keyboard where the mod key is between the space and alt key, and a wireless keyboard where the mod and alt keys are swapped.
I have a config file for each device, where the keymaps are the same. The two files differ only by their modmap.
Being able to import files would allow me to factor out the common keymaps instead of having to apply the same changes to both files.

ChingChang9:
Resolves #202 (sorry for the delay! and thanks for the patience 😅).

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -17,7 +17,7 @@ use keymap::Keymap;
 use modmap::Modmap;
 use nix::sys::inotify::{AddWatchFlags, InitFlags, Inotify};
 use serde::{Deserialize, Deserializer};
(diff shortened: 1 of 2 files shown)


k0kubun (approved):
Great job!

Issue #298, opened 2023-05-30
Feature request: map one key to many mods


hab25:
e.g., support
```json
{
  "modmap": [
    {
      "remap": {
        "rightalt": "control-leftalt-shift-super"
      }
    }
  ]
}
```

---
Use case: I have a lot of shortcuts using `rightalt` (which almost all my applications understand as `AltGr`), but [Emacs cannot (directly) use `AltGr` as a modifier](https://github.com/emacs-mirror/emacs/blob/31a80f61ec03bcbb79720c0dc640272aba160865/doc/emacs/custom.texi#L2030). As a workaround, I want to do the mapping in the example above. Currently, I can achieve this in practice by mapping each of the shortcuts manually, e.g. 
```json
{
  "keymap": [
    {
      "application": {
        "only": "Emacs"
      },
      "remap": {
        "rightalt-a": "control-leftalt-shift-super-a",
        "rightalt-b": "control-leftalt-shift-super-b",
        "rightalt-c": "control-leftalt-shift-super-c"
      }
    }
  ]
}
```
As I have many more than 3 shortcuts using `rightalt`, doing this is much more repetitive and error-prone than what the feature would enable.

k0kubun:
Sounds good. I slightly prefer `"rightalt": ["control", "leftalt", "shift", "super"]` over `"rightalt": "control-leftalt-shift-super"`, but overall this idea makes sense.

I personally don't have a use case for this, but I'd be open to merging a pull request that adds this feature if anybody is interested in making one.

jixiuf:
I remap my  spacebar with  [dual-function-keys](https://gitlab.com/interception/linux/plugins/dual-function-keys)  
1. tap for spacebar
2. hold for  `super+control+alt`

k0kubun:
@jixiuf I've seen you recommended dual-function-keys in https://github.com/k0kubun/xremap/issues/290#issuecomment-1537276322 too, but I'm not sure how it helps xremap solve these issues. Do you use xremap and dual-function-keys together? If so, what is the missing feature that prevents you from using a single key remapper?

k0kubun:
Thanks for your reply. I created another thread https://github.com/k0kubun/xremap/issues/306 for this discussion to avoid deviating this thread too much. Let's keep dual-function-keys -related conversations there instead of hinting that we need the feature in multiple other issues.

k0kubun:
@jixiuf implemented this :tada: https://github.com/k0kubun/xremap/pull/339

PR #339, opened 2023-09-06
modmap held/alone support multi key

### LINKED ISSUE #306: Dual function keys
Re: https://github.com/k0kubun/xremap/issues/290#issuecomment-1537276322, https://github.com/k0kubun/xremap/issues/298#issuecomment-1581965311

@jixiuf combines both xremap and [dual-function-keys](https://gitlab.com/interception/linux/plugins/dual-function-keys) to achieve:

1. tap for spacebar
2. hold for `super+control`

But we might be able to extend xremap's `held`/`alone` to support this.

---

As to macOS compatibility, we could add a script to convert an xremap YAML file into a Karabiner-Element config JSON file. It should, however, live in a separate repository, e.g. xremap/xremap-karabiner-elements.

jixiuf:
close #306 #298 
demo:
```
modmap:
  - name: Global
    remap:
      space:
        held: [ctrl_l,super_l]
        alone: space
        alone_timeout_millis: 1000 # Optional
      super_r :
        held: super_l
        alone: [ctrl_l,space]
        alone_timeout_millis: 1000 # Optional
```

### DIFF
--- a/src/config/modmap_action.rs
+++ b/src/config/modmap_action.rs
@@ -4,7 +4,10 @@ use serde::{Deserialize, Deserializer};
 use serde_with::{serde_as, DurationMilliSeconds};
 use std::time::Duration;
 
-use super::keymap_action::{Actions, KeymapAction};
+use super::{
+    deserialize_virtual_modifiers,
+    keymap_action::{Actions, KeymapAction},
+};
 
 // Values in `modmap.remap`
 #[derive(Clone, Debug, Deserialize)]
@@ -19,10 +22,8 @@ pub enum ModmapAction {
 #[serde_as]
 #[derive(Clone, Debug, Deserialize)]
 pub struct MultiPurposeKey {
-    #[serde(deserialize_with = "deserialize_key")]
-    pub held: Key,
-    #[serde(deserialize_with = "deserialize_key")]
-    pub alone: Key,
+    pub held: Keys,
+    pub alone: Keys,
     #[serde_as(as = "DurationMilliSeconds")]
     #[serde(default = "default_alone_timeout", rename = "alone_timeout_millis")]
     pub alone_timeout: Duration,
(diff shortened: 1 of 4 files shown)


k0kubun (approved):
Thanks!

src/config/modmap_action.rs
```diff
@@ -19,10 +19,8 @@ pub enum ModmapAction {
 #[serde_as]
 #[derive(Clone, Debug, Deserialize)]
 pub struct MultiPurposeKey {
-    #[serde(deserialize_with = "deserialize_key")]
-    pub held: Key,
-    #[serde(deserialize_with = "deserialize_key")]
-    pub alone: Key,
+    pub held: Keys,
```
k0kubun, line 22: Could you also accept a non-vector `Key` for backward compatibility? For reference, we do something similar in multiple places, e.g. `Application::only`. They accept a string but also a list of strings in the same place. Many people could still use this feature without specifying multiple keys here, so that would be useful.

jixiuf: Keys is an emun, it can be a non-vector 'Key'
```
pub enum Keys {
    #[serde(deserialize_with = "deserialize_key")]
    Key(Key),
    #[serde(deserialize_with = "deserialize_virtual_modifiers")]
    Keys(Vec<Key>),
}
```

k0kubun: Oh okay, sorry about that. I got confused by changes in the documentation and tests.

src/config/tests.rs
```diff
@@ -56,12 +56,12 @@ fn test_modmap_multi_purpose_key() {
...
-            held: Shift_L
-            alone: Space
+            held: [Shift_L]
+            alone: [Space]
       - remap:
           Muhenkan:
-            held: Alt_L
-            alone: Muhenkan
+            held: [Alt_L]
+            alone: [Muhenkan]
```
k0kubun, line 64: Can you also leave a test that parses the non-vector case?

jixiuf: done

k0kubun: I meant a test case without `[` and `]`. Could you just add a new test instead of rewriting an existing test?

jixiuf: done

k0kubun: Thanks!

k0kubun:
One last thing, could you run `cargo fmt` locally and push the change so that the CI becomes green?

jixiuf:
I have tried that, but I am not familiar with rust and I don’t know how to make `cargo fmt` work. The result when I execute cargo fmt is as follows. So can you fix that for me .
图片

k0kubun:
OK, can you apply the diff that the CI is showing then?

jixiuf:
> OK, can you apply the diff that the CI is showing then?

done

k0kubun (approved):
Thanks a lot!

PR #346, opened 2023-09-28
add option for unique id

Buttars:
This is a rough draft state. Learned a lot along the way about rust. Please let me know if there's anything I can improve on.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -1,0 +1,60 @@
+use clap::Parser;
+use clap_complete::Shell;
+use std::path::PathBuf;
+
+use super::device::InputDevice;
+use super::watch_targets::WatchTargets;
+
+#[derive(Parser, Debug)]
+#[clap(version)]
+#[allow(non_snake_case)]
+pub struct Args {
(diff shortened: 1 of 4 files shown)


Buttars:
One concern I have is the refactoring of Args and WatchTargets as well as the several methods I had to change to public to get this to work. 

I'm wondering if the already public get_input_devices method should be refactored to only return devices and exclude all the logging that way it can be used throughout the application without spamming the terminal.

Or if another public method should be exposed to fetch and filter for devices similarly.

Buttars:
I'm happy to continue refining this PR. The reason for the breaking out of the new files was because of the need to expose the Args and WatchTargets structs. Is there a way to expose/import those without moving them out of the files?

k0kubun:
Oh okay. I'm sure it's possible, but let me have a look again tonight to share how it can be done.

k0kubun:
Ugh, I accidentally pushed this change to the actual master since your branch had the same name. I'll reopen another one.

PR #369, opened 2023-11-08
Add Alt-backspace config into example/emacs.yml

zw963:
Add emacs mapping for `Kill word backward` use `Alt-backspace`, and exclude this config for emacs itself too.

Thanks

### DIFF
--- a/example/emacs.yml
+++ b/example/emacs.yml
@@ -36,6 +36,8 @@ keymap:
       C-k: [Shift-end, C-x, { set_mark: false }]
       # Kill word backward
       Alt-backspace: [C-backspace, {set_mark: false}]
+      # set mark next word continuously.
+      C-M-space: [C-Shift-right, {set_mark: true}]
       # Undo
       C-slash: [C-z, { set_mark: false }]
       C-Shift-ro: C-z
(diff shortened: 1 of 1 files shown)


example/emacs.yml
```diff
@@ -32,6 +34,8 @@ keymap:
       M-d: [C-delete, { set_mark: false }]
       # Kill line
       C-k: [Shift-end, C-x, { set_mark: false }]
+      # Kill word backward
+      Alt-backspace: [C-backspace, C-x, {set_mark: false}]
```
k0kubun, line 38: In my environment (Ubuntu 22.04, X11 GNOME), `C-backspace` already deletes a word, so `C-x` seems to do nothing. Why did you add `C-x` here?

zw963: Sorry, my fault, i modified config from the copy of above `C-k:`  :smile:

I added a C-M-Space config too, please check.

zw963:
Hi, i do propose change `C-s: f3` into `C-s: C-f`, because after user search a keyword use F3 at the first time, then pressing F3(C-s) again, which will search next previous keyword instead of popup the search menu, which is not we want.

Thanks

k0kubun (approved):

Issue #201, opened 2022-11-14
device-specific remapping


tgy:
hi! thanks for this awesome software! really useful

is there a way to remap only for a specific device?

i have an external typematrix keyboard and there are some keys i want to remap, but only for that keyboard

thx!

k0kubun:
Hi. Yes, there is. I'm not in front of my machine and can't show you how, but there's --device option. Please have a look at "xremap --help", see output of xremap boot, and specify a device name according to the output.

tgy:
thanks for your quick reply! really helpful.

we could add something about that to the README

k0kubun:
Sounds good. Can you make it a pull request?

eeshugerman:
What if I want to remap different things for different devices? Should I run multiple xremap processes with different config files?

k0kubun:
I think we should introduce `device.only` and `device.not` for it, in the same way as `application.only` and `application.not`. Would you be interested in filing a pull request to add that feature?

eeshugerman:
Sweet, yeah, that was going to be my next question! I'll take a stab at it

dtcooper:
Any update on `device.only` and `device.not`? :) Cheers and thanks

eeshugerman:
I haven't made much progress yet. I still intend to make it happen, but can't say when. If someone beats me to it that's fine too :)

dtcooper:
Your branch looks pretty good to me, @eeshugerman! Is it documented? How do I use it?

eeshugerman:
Well I just got it to compile last weekend. Haven't had a chance to test it at all yet so I'd be shocked if it works, but feel free to try it out!

I haven't added documentation yet, but it should work just like `application.only`/`application.not`, which is documented in the README. The strings you provide to identify a device should be the same as those passed to the `--device` flag, which doesn't appear to be documented, but there are some hints [here](https://github.com/k0kubun/xremap/issues/201#issuecomment-1313925478) and [here](https://github.com/k0kubun/xremap/blob/a73ad9cc06e394e595900fe9474893c213ffa3e7/src/device.rs#L232-L245).

eeshugerman:
Update: Just tried it out and it seems to be working. Here's the branch if anyone's interested: https://github.com/eeshugerman/xremap/tree/device-aware-config

Still some work to do on tests and such to get this to a point where it's mergable but I should have PR up before too long.

eeshugerman:
[...]

PR #380, opened 2023-11-27
Support device-specific remapping in config

eeshugerman:
This patch adds support for `device.only` and `device.not`, akin to `application`, as suggested [here](https://github.com/k0kubun/xremap/issues/201#issuecomment-1414135214).

The matching logic from `--device` is reused, so devices may be specified by name, path, etc. For simplicity, regex matching is not implemented here. (I would not object to it being added later if desired.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@
 * Remap a key sequence as well. You could do something like Emacs's `C-x C-c`.
 * Remap a key to two different keys depending on whether it's pressed alone or held.
 * Application-specific remapping. Even if it's not supported by your application, xremap can.
+* Device-specific remapping.
 * Automatically remap newly connected devices by starting xremap with `--watch`.
 * Support [Emacs-like key remapping](example/emacs.yml), including the mark mode.
 * Trigger commands on key press/release events.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -350,6 +351,31 @@ keymap:
 
 Note how Alt-f and Alt-b work in all apps, but the definition of Alt-f is slightly different in LibreOffice Writer. When that app is active, the first definition overrides the second definition; but for any other app, only the second definition is found. This is because xremap uses the first matching definition that it finds.
 
+### device
```
k0kubun, line 354: :+1: Nice documentation. For better discoverability, similar to `application`, could you mention this under `modmap` and `keymap` as well? (like the `application: # Optional` part)

k0kubun:
Looking pretty good. I left only a minor comment. It'd be nice if you have a chance to look at it, but I'll merge & release this tomorrow regardless.

README.md
eeshugerman: Will do!

eeshugerman: Not sure I'll get to this today though, probably tomorrow. Feel free to merge today or wait, either is fine by me

k0kubun: No worries, I'll just release the current version first :+1:

Issue #401, opened 2024-01-03
can deny_unknown_fields in config parser (serde_yaml) be removed?


Pieterv24:
I've noticed the use of `#[serde(deny_unknown_fields)]`on the structs that are used to parse the config.

Is there any usecase/reason for this?

If not, I would like to propose removing these attributes.
This would allow the use of Anchors and Aliases (as per the [yml spec](https://yaml.org/spec/1.2.2/#3222-anchors-and-aliases))

When using the following config
```yaml
terminals: &terminals
  - kitty
  - xterm
  - gnome-terminal
  - etc
 
modmap:
  - name: Terminals
    application:
      only: *terminals
    remap:
      Alt_L: Ctrl_R
```

An error is thrown by the serde deserialiser because `terminals` is an unknown field.

If there is no usecase for these attributes, I would like to propose their removal.
(I could create a PR for this if needed)

k0kubun:
The use case is validation. If you make a typo, you wouldn't notice without it. ref: https://github.com/k0kubun/xremap/issues/359

I'm open to having such a capability though. WDYT about adding a new top-level field like `shared:` that allows unknown fields and putting `terminals:` under that while denying unknown fields for the rest of the places?

Pieterv24:
> The use case is validation. If you make a typo, you wouldn't notice without it. ref: #359

I see, I understand your point.

> I'm open to having such a capability though. WDYT about adding a new top-level field like `shared:` that allows unknown fields and putting `terminals:` under that while denying unknown fields for the rest of the places?

I´d say that'd be a good compromise between validation and the possibility to use anchors and aliases.

I'll see if I can get that to work.

PR #402, opened 2024-01-03
Added shared data field and tests

Pieterv24:
As discussed in #401 

I've added the `shared` field to the config struct for use with out of scope data.
This data can be used with Aliases and References.

The datatype is `ignoreAny` which is a way of discarding data in the deserializer.

Please let me know if any additions or changes are needed.


k0kubun (approved):
Thanks!

PR #404, opened 2024-01-07
Add TOML support to xremap

mattamend:
Apologies, forgot to commit my changes to Cargo.toml (and Cargo.lock). This is my second try, hopefully this passes :-)

I added TOML support in config::load_configs. First, it will check if the config file is valid YAML, then it will check for a valid TOML config file.

Tests are also updated accordingly.

### DIFF
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -53,25 +53,38 @@ pub struct Config {
     pub keymap_table: HashMap<Key, Vec<KeymapEntry>>,
 }
 
+enum ConfigFiletype {
+    Yaml,
+    Toml,
+}
+
+fn get_file_ext(filename: &PathBuf) -> ConfigFiletype {
+    match filename.extension() {
+        Some(f) => {
+            if f.to_str().unwrap_or("").to_lowercase() == "toml" {
+                ConfigFiletype::Toml
+            } else {
+                ConfigFiletype::Yaml
+            }
+        },
+        _ => ConfigFiletype::Yaml,
+    }
+}
+
 pub fn load_configs(filenames: &Vec<PathBuf>) -> Result<Config, Box<dyn error::Error>> {
     // Assumes filenames is non-empty
     let config_contents = fs::read_to_string(&filenames[0])?;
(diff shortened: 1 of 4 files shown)


src/config/tests.rs
```diff
@@ -1,10 +1,284 @@
 use crate::Config;
 use indoc::indoc;
-use serde_yaml::Error;
+
+extern crate serde_yaml;
+extern crate toml;
+
+#[test]
+fn test_toml_modmap_basic() {
```
k0kubun, line 8: Could you put newer tests after the ones that currently exist? I want to see YAML tests first since the syntax is still primarily designed for YAML.

src/config/mod.rs
```diff
@@ -54,12 +55,25 @@ pub struct Config {
...
     // Assumes filenames is non-empty
-    let yaml = fs::read_to_string(&filenames[0])?;
-    let mut config: Config = serde_yaml::from_str(&yaml)?;
+    let config_contents = fs::read_to_string(&filenames[0])?;
+    let mut is_yaml = true;
+
+    let mut config: Config = {
+        match serde_yaml::from_str(&config_contents) {
+            Ok(c) => c,
+            Err(_) => { is_yaml = false; toml::from_str(&config_contents)? }
```
k0kubun, line 64: When you use a YAML file and there's a syntax error in it, this would report a syntax error as a TOML file. Can we just look at the filename's extension and use TOML only when it ends with `.toml` and use YAML if otherwise (because it'd be backward-compatible when there's no extension and things like `.json` is also a valid YAML file)?

k0kubun:
I'm open to adding this feature since "Language-agnostic" has been the core of xremap's design from the beginning. But I want it to switch the config parser based on the filename's extension https://github.com/k0kubun/xremap/pull/404#discussion_r1444159724, which is far more scalable when we start adding many other languages.

mattamend:
This should be fixed to use filename extensions now. `enum ConfigFileType` should improve scalability in terms of new language support.

k0kubun (approved):
:+1:

Issue #433, opened 2024-03-08
No longer working with Plasma 6 Wayland


HanPrower:
I've had xremap setup and working on Arch Linux before on Plasma Wayland 5.2x so it's configured correctly, but it's just existed for a while without me needing it, so it's possible it stopped working earlier...

So, xremap still seems to work in-so-much as it runs, can find the devices, even remap keys if the only application is set to `not`, but then when it attempts to connect the KDEClient to detect the Active Window it to fails.

`Mar 08 18:55:40 smokey systemd-logind[616]: Watching system buttons on /dev/input/event22 (xremap)`

2 minutes later...

`Mar 08 18:57:54 smokey xremap[28122]: application-client: KDE (supported: false)`

It seems to be failing around [Line 121 of kde_client.rs](https://github.com/k0kubun/xremap/blob/master/src/client/kde_client.rs#L121). I guess it's failing to create `com.k0kubun.Xremap`? I don't see it in QDBusViewer... but I don't know exactly what should be happening.

Happy to debug and try to fix it with guidance :)

HanPrower:
Thanks, rebuilding with those changes does indeed let the client start and report window changes.

Although for some reason it can fail multiple times until the `application-client` starts successfully. Not sure if that's a Plasma 6 issue or a me issue...

k0kubun:
Hmm. Should I push that change then? It seems hard to support multiple KDE versions in that case though :confused: 

@N4tus As the author of https://github.com/k0kubun/xremap/pull/264, WDYT?

Ahrotahn:
If you don't care about adding another release target then an alternative option for supporting both would be to split the `kde` feature into separate `kde5` and `kde6` features.  Then the changes can be gated by the feature used.

N4tus:
First of, thank you @Ahrotahn for figuring out what api changed. This would have taken me countless hours to discover myself, and I truly appreciate it.

My original idea was to have another back-end specifically for plasma 6. So that a kde and kde6 back-end could live side by side and the user would have to choose the one they need. But if we only need to change the object-path and load a different script/patch the existing script depending on the version, my idea might be not as good.

[...]

Issue #393, opened 2023-12-21
application-specific modmap not working for KDE wayland


ydzhou:
modmap applies to all applications and ignore my application specific config.

Here is the example config:
```
modmap:
  - name: OSX style binding
    remap:
      Win_L: Ctrl_L
      Win_R: Ctrl_R
    application:
      not: [google-chrome, kitty]
```

Google-chrome and Kitty still get applied with Win key swap

- xremap version: 0.8.12
- KDE: 5.27.10
- wayland: 1.22.0
- OS: Linux 6.6.7

k0kubun:
Can you run `RUST_LOG=debug xremap ...`, press the combination you want to remap, and report the entire output (including the beginning) of the command?

ydzhou:
Expected behavior is that when I press WIN key, it should still be WIN key for kitty. But xremap ignore my not rule and apply modmap to make WIN key to CTRL key

Output
```
Selected keyboards automatically since --device options weren't specified:
------------------------------------------------------------------------------
/dev/input/event13: Microsoft Microsoft® Nano Transceiver v2.1
/dev/input/event16: Microsoft Microsoft® Nano Transceiver v2.1 Consumer Control
/dev/input/event19: SteelSeries SteelSeries Sensei 310 eSports Mouse Keyboard
/dev/input/event20: "Drop" "ALT"
/dev/input/event24: "Drop" "ALT" Keyboard
------------------------------------------------------------------------------
[2023-12-21T19:49:15Z DEBUG xremap::event_handler] => 0: KEY_ENTER
[2023-12-21T19:49:15Z DEBUG xremap::action_dispatcher] 0: KEY_ENTER
[2023-12-21T19:49:25Z DEBUG xremap::event_handler] => 1: KEY_LEFTMETA
application-client: KDE (supported: true)
[2023-12-21T19:49:25Z DEBUG xremap::action_dispatcher] 1: KEY_LEFTCTRL
[2023-12-21T19:49:25Z DEBUG xremap::event_handler] => 2: KEY_LEFTMETA
[2023-12-21T19:49:25Z DEBUG xremap::action_dispatcher] 2: KEY_LEFTCTRL
[2023-12-21T19:49:25Z DEBUG xremap::event_handler] => 1: KEY_T
[2023-12-21T19:49:25Z DEBUG xremap::action_dispatcher] 1: KEY_T
^T[2023-12-21T19:49:25Z DEBUG xremap::event_handler] => 0: KEY_T
[...]

PR #437, opened 2024-03-15
Handle new KWin API for kde-backend.

N4tus:
I have put up the required changes that should be enough to support both versions. BUT I don't have a newer KDE installation and I need someone else to test it. (That is why it's currently WIP). It does work for me at least.

The Kwin-Script dispatches based on the comment that was linked in the issue and for starting the script I just try out both paths.

#433

### DIFF
--- a/src/client/kde_client.rs
+++ b/src/client/kde_client.rs
@@ -29,14 +29,14 @@ impl Drop for KwinScriptTempFile {
 }
 
 trait KWinScripting {
-    fn load_script(&self, path: &Path) -> Result<String, ConnectionError>;
+    fn load_script(&self, path: &Path) -> Result<i32, ConnectionError>;
     fn unload_script(&self) -> Result<bool, ConnectionError>;
-    fn start_script(&self, script_obj_path: &str) -> Result<(), ConnectionError>;
+    fn start_script(&self, script_obj_id: i32) -> Result<(), ConnectionError>;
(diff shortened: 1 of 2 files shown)


k0kubun:
By the way, as I raised at https://github.com/k0kubun/xremap/issues/393#issuecomment-1879834308 before, would you be interested in being a maintainer for the KDE client?

Ahrotahn:
These changes are functional when I tested them on my plasma 6 system.

N4tus:
Since @Ahrotahn confirmed that it works on plasma 6, I think we can merge it.  Although it also would be interesting to know if #393 was also fixed with this PR. But even if not, supporting kde6 is still an improvement.

Also to answer your Question @k0kubun: You can count me as maintainer for the KDE-Client.

k0kubun:
> Also to answer your Question @k0kubun: You can count me as maintainer for the KDE-Client.

Thanks! I transferred this repository to the xremap org and invited you to it to give you write access to this repository. Please accept the invitation when you have a chance.

k0kubun:
Thanks for working on this. 

By the way, the release is fully automated using git tags, so you can cut a release too. Write up a CHANGELOG entry, update versions in Cargo.toml/lock, make a commit, tag it with `vX.Y.Z`, and push the tag and branch.

```bash
git commit -m "Version 0.8.16"
git tag v0.8.16
git push origin master && git push origin --tags
```

PR #439, opened 2024-03-18
Clearer sudo-less instructions in README

EricDriussi:
Hi there!
Thanks a lot for this awesome piece of software, it's been quite helpful and painless.

One thing I found a bit odd during the setup was the fact that the instructions regarding sudo-less usage are hidden behind a fold and some distro specific commands seem to be copy pasted around.

Since using sudo is discouraged when avoidable and the instructions for arch and debian are more or less the same (and I also got it to work on fedora by following them), I gave the readme a little refactor to hopefully give a clear guide to distro-agnostic sudo-less setup for new users.

If there's anything you find odd just let me know and I'll update the PR :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,77 +62,25 @@ Then start the `xremap` daemon by running:
 sudo xremap config.yml
 ```
 
-(You will need to leave it running for your mappings to take effect.)
+To use it without sudo (recommended) run the following commands:
 
-<details>
-<summary>If you want to run xremap without sudo, click here.</summary>
-
-### Running xremap without sudo
-
-To do so, your normal user should be able to use `evdev` and `uinput` without sudo.
-In Ubuntu, this can be configured by running the following commands and rebooting your machine.
-
-```bash
+```sh
 sudo gpasswd -a YOUR_USER input
-echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/input.rules
+echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/90-input.rules
+echo 'uinput' | sudo tee /etc/modules-load.d/uinput.conf
 ```
 
-#### Arch Linux
+These changes will take effect after a reboot, to load them for the current session use:
 
-The following can be used on Arch.
-
-```bash
-lsmod | grep uinput
-```
-If this module is not loaded, add to `/etc/modules-load.d/uinput.conf`:
-```bash
-uinput
-```
-Then add udev rule.
-
-```bash
-echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/99-input.rules
-```
-
-Then reboot the machine.
-
-#### Debian
-Make sure `uinput` is loaded same as in Arch:
-```
-lsmod | grep uinput
-```
-If it shows up empty:
-```bash
-echo uinput | sudo tee /etc/modules-load.d/uinput.conf
-```
-Add your user to the `input` group and add the same udev rule as in Ubuntu:
-```bash
-sudo gpasswd -a YOUR_USER input
-echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/input.rules
-```
-Reboot the machine afterwards or try:
-```bash
+```sh
 sudo modprobe uinput
 sudo udevadm control --reload-rules && sudo udevadm trigger
 ```
 
-#### Other platforms
-
-In other platforms, you might need to create an `input` group first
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,85 +62,39 @@ Then start the `xremap` daemon by running:
...
-If this module is not loaded, add to `/etc/modules-load.d/uinput.conf`:
-```bash
-uinput
-```
-Then add udev rule.
-
-```bash
-echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/99-input.rules
+echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/90-input.rules
+echo 'uinput' | sudo tee /etc/modules-load.d/uinput.conf
```
k0kubun, line 70: My environment doesn't need this. This should go to platform-specific sections.

README.md
```diff
@@ -62,85 +62,39 @@ Then start the `xremap` daemon by running:
 sudo xremap config.yml
 ```
 
-(You will need to leave it running for your mappings to take effect.)
+To use it without sudo (recommended) run the following commands:
```
k0kubun, line 65: It's a lot of work to make it work, so I actually don't recommend it. It's still often necessary to make `application`-specific remappings work in many platforms. Let's remove "(recommended)" and indicate that it's needed for `application` filters.

README.md
```diff
@@ -62,85 +62,39 @@ Then start the `xremap` daemon by running:
...
 
-<details>
```
k0kubun, line 67: If you unfold this, please at least have a separate sub-section, i.e. `###`. Having two distinct options in the same section seems confusing.

README.md
EricDriussi: I'm not sure about this: I just tried a fresh Ubuntu Desktop 22.04 install on a VM and indeed without this setup xremap produces a `Permission denied` error.
Are you sure this is not needed for Ubuntu?

k0kubun: Here's my Ubuntu 22.04 environment right now:

```
$ ls -la /etc/modules-load.d/
total 20
drwxr-xr-x   2 root root  4096 Jan 28 19:53 .
drwxr-xr-x 147 root root 12288 Mar 21 08:54 ..
-rw-r--r--   1 root root   119 Apr 11  2022 cups-filters.conf
lrwxrwxrwx   1 root root    10 Nov 21 12:57 modules.conf -> ../modules

$ cat /etc/modules-load.d/cups-filters.conf
# Parallel printer driver modules loading for cups
# LOAD_LP_MODULE was 'yes' in /etc/default/cups
lp
ppdev
parport_pc

$ cat /etc/modules-load.d/modules.conf
# /etc/modules: kernel modules to load at boot time.
#
# This file contains the names of kernel modules that should be loaded
# at boot time, one per line. Lines beginning with "#" are ignored.
```

As I've set up Ubuntu environments from scratch with these instructions multiple times myself, I'm pretty sure it has not been necessary at least for Ubuntu 20.04 and 22.04.

k0kubun: In other words, Ubuntu loads uinput by default. This is not the case for highly customizable distributions like Arch Linux, but it sounds very normal that distributions like Ubuntu have this kind of thing out of the box.

EricDriussi: Understood, hopefully my latest commit makes this more distro-specific

README.md
```diff
@@ -58,90 +58,57 @@ Write [a config file](#Configuration) directly, or generate it with
 
 Then start the `xremap` daemon by running:
 
-```
+```sh
 sudo xremap config.yml
 ```
 
-(You will need to leave it running for your mappings to take effect.)
+You might need to run it with sudo for application-specific remaps to work in
```
k0kubun, line 65: question: What does this mean? I think it's generally easier to make application-specific remaps work without sudo. Sudo makes everything but application-specific remaps easier.

README.md
```diff
@@ -58,90 +58,57 @@ Write [a config file](#Configuration) directly, or generate it with
...
+To use it without sudo just add your user to the `input` group:
 
-```bash
+```sh
 sudo gpasswd -a YOUR_USER input
-echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/input.rules
 ```
 
-#### Arch Linux
+While this is enough for some distros (Ubuntu), others (Fedora, Arch, Debian)
```
k0kubun, line 78: > While this is enough for some distros (Ubuntu)

It's not. As indicated by the current README, `echo 'KERNEL=="uinput", GROUP="input", TAG+="uaccess"' | sudo tee /etc/udev/rules.d/input.rules` is necessary, at least for Ubuntu.

EricDriussi:
Sorry, I don't think we are understanding each other. I don't sense much interest on actually updating the readme either so I'll just close the PR for now.
Thanks for your time.

PR #460, opened 2024-04-27
set_mode and launch command at same time

blob42:
This allows to run custom commands when a mode is entered


k0kubun:
Is the following config not sufficient?

```yml
keymap:
  - remap: # Required
      Esc: [{set_mode: normal}, {launch: ["wmctrl", "-x", "-a", "code.Code"]}]
```

blob42:
@k0kubun I had no idea it was possible. Yes that's enough. It would be great if the docs mention that a keymap can be a list of any keymap action.

PR #462, opened 2024-04-28
better to send no key for skip event

blob42:
I didn't think much about it on my first patch but when skip_key_event
is set the KEY_UNKNOWN is repeated. Better to send no key at all.


k0kubun:
Agreed, thanks!

PR #478, opened 2024-05-18
enable wheel by default

### LINKED ISSUE #260: Mouse wheel stops working with v0.8.2
After upgrading to v0.8.2 my mouse wheel doesn't work, presumably due to 2603e89999a25f6d65d2bce388ddb6f387101416. Is this intentional and/or workaroundable? 

This is using sway (--features sway).

KeenS:
close #260


k0kubun (approved):
Thanks!

Issue #523, opened 2024-09-03
Doesn't work with Zed


DidacGit:
Hey, first of all, amazing software!

It seems like xremap can't retrieve the name of the [Zed](https://zed.dev/) window. For example, using this simple config:

```yml
keymap:
  - name: Emacs inspired
    application:
      only: ["dev.zed.Zed"]
    remap:
      C-f: { with_mark: right }
```
When I run it with xremap, this is the output, of typing `C-f` in Mousepad and then typing it on Zed:
```
application-client: X11 (supported: true)
application: mousepad.Mousepad
Reconnecting to X11 due to error: X11 error X11Error { error_kind: Window, error_code: 3, sequence: 20, bad_value: 0, minor_opcode: 0, major_opcode: 20, extension_name: None, request_name: Some("GetProperty") }
Reconnecting to X11 due to error: X11 error X11Error { error_kind: Window, error_code: 3, sequence: 2, bad_value: 0, minor_opcode: 0, major_opcode: 15, extension_name: None, request_name: Some("QueryTree") }
```
I got the `dev.zed.Zed` name by using `xprop`, by the way.
Any idea of what could be wrong? I'm on Debian 12, XFCE (X11). I have Zed installed locally, normally, so no Flatpak nor anything weird.
Thanks in advance!

wistoft:
I could confirm the problem as you described it, and I've made a PR with a fix. It would be nice, if you could try it out, and see if it fixes your problem.

k0kubun:
Released the patch as v0.10.1. Thank you!

DidacGit:
Thank you guys! Now it works perfectly.

PR #525, opened 2024-09-14
X11: Handle WM_CLASS reply without terminating null-byte

wistoft:
The is issue #523 was due x11_client not handling the WM_CLASS response from xserver gracefully. The server can send a terminating null-byte in the response, and it can omit the terminating null-byte. For zed it apparently omits it.


k0kubun (approved):
Thanks!

PR #544, opened 2024-10-30
Allow keymap actions with no action

### LINKED ISSUE #297: Recommended way to disable a key sequence?
It's probably worth mentioning the answer to this question in the README.

Note:
I'm currently achieving this by using the [reserved key](https://github.com/emberian/evdev/blob/1d020f11b283b0648427a2844b6b980f1a268221/src/scancodes.rs#LL25C6-L25C6).
The use case is to learn a new keymap faster when the key you want to disable should not be used for anything. For example: 

---
I want to start using `C-l` to close a google chrome tab, I don't want to use use `C-w` for anything, and currently, I'm using `C-w` to close a google chrome tab (this a google chrome default and cannot be disabled). I can do
```json
{
  "keymap": [
    {
      "application": {
        "only": "Google-chrome"
      },
      "exact_match": true,
      "remap": {
        "c-w": "reserved",
        "c-l": "c-w"
      }
    }
  ]
}
```
This way, `C-w` will start not doing anything which will help me break the habit of using it to close tabs. 

---
Q: Why not use two keybinds to do the same thing?
A: Spreading your habit among multiple keybinds weakens the muscle memory for any single keybind. Further, it makes the event that one of the keybinds is changed into some other command in the future easier to learn. More abstractly, one thing is generally simpler than multiple.

wistoft:
Fix #297

### Feature
Suppress the default action with either of these to remaps:

```yml
keymap:
    - remap:
        f11: []
        f12: null
```
`null` is syntactic sugar for an empty vector.

### Commits
- Add tests of existing functionality.
  - I have added tests to illustrate the changes this commit will cause.
- Allow keymap actions to have no action
  - The function is_remap is changed to not consider an empty keymap action as a nested remap.
  - `Actions` struct for parsing config files has a new enum `NoAction` that turns into an empty vec. This allows:
    1. `null` for keymap action
    2. `null` for actions in nested remaps.
    3. `null`  for press/release key actions.

### Breaking changes
I'm not sure these are breaking changes, as they are quite strange use cases, but I will call them out, so they can be considered.

#### Change 1

[test_mixing_keypress_and_remap_in_keymap_action](https://github.com/wistoft/xremap/commit/c6d874c1edc60d32630eb4ca23c52accd705ef60#diff-a0262fea108af44a98cc49e5eb72641963dd816513f2d0a22eb738fcfed55ebeL686) shows that the following config will change behavior:
```yml
keymap:
    - remap:
        f12: []
    - remap:
        f12:
        - remap:
            a: b
```
`f12` would previously activate the nested remap, because the `f12: []` was effectively ignored. Now `f12: []` means emit nothing, so the search stops there.

I think this is acceptable, because `f12: []` had no effect before. Why would user's have it in their config files.

#### Change 2

[test_no_keymap_action](https://github.com/wistoft/xremap/commit/c6d874c1edc60d32630eb4ca23c52accd705ef60#diff-a0262fea108af44a98cc49e5eb72641963dd816513f2d0a22eb738fcfed55ebeL728) shows that the following config will change behavior:

```yml
keymap:
    - remap:
        f12: []
```
`f12` would previously emitted `f12`, as if there was no configuration. Now it will suppress `f12`.

I think this is acceptable because the config was a no-op before, so user's could just remove it. And why would they have it in the first place.

### DIFF
--- a/src/config/keymap_action.rs
+++ b/src/config/keymap_action.rs
@@ -135,3 +135,50 @@ impl Actions {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::KeymapAction;
+    use crate::config::key_press::KeyPress;
+    use crate::config::key_press::Modifier;
+    use evdev::Key;
+
+    #[test]
+    fn test_keypress_action() {
+        test_yaml_parsing_key_press(
+            "c-x",
+            KeyPress {
+                key: Key::KEY_X,
+                modifiers: vec![Modifier::Control],
+            },
+        );
+    }
+
+    #[test]
+    fn test_launch_action() {
+        test_yaml_parsing_key_launch("{launch: []}", vec![]);
+        test_yaml_parsing_key_launch("{launch: [\"bla\"]}", vec!["bla".into()]);
+    }
+
+    //
+    // util
+    //
+
+    fn test_yaml_parsing_key_press(yaml: &str, expected: KeyPress) {
+        match serde_yaml::from_str(yaml).unwrap() {
+            KeymapAction::KeyPress(keyp) => {
+                assert_eq!(keyp, expected);
+            }
+            _ => panic!("unexpected type"),
+        }
+    }
+
+    fn test_yaml_parsing_key_launch(yaml: &str, expected: Vec<String>) {
+        match serde_yaml::from_str(yaml).unwrap() {
+            KeymapAction::Launch(vect) => {
+                assert_eq!(vect, expected);
+            }
+            _ => panic!("unexpected type"),
+        }
+    }
+}
--- a/src/config/keymap_action.rs
+++ b/src/config/keymap_action.rs
@@ -123,6 +123,8 @@ where
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
 pub enum Actions {
+    // Allows keychords to map to null, which means no actions.
+    NoAction,
     Action(KeymapAction),
     Actions(Vec<KeymapAction>),
 }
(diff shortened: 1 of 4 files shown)


wistoft:
This PR fixes: https://github.com/xremap/xremap/issues/297

wistoft:
Should I make this without the breaking change. I could leave `f11: []` as it is, and only make it for `f11: null`. That can be done non-breaking with a new KeymodAction.

src/event_handler.rs
```diff
@@ -697,6 +697,10 @@ impl EventHandler {
 }
 
 fn is_remap(actions: &Vec<KeymapAction>) -> bool {
+    if actions.len() == 0 {
```
k0kubun, line 700: Makes sense, but the meaning of this code wasn't immediately clear to me. Could you leave a comment above this to clarify how this works and what this is for?

src/config/keymap_action.rs
```diff
@@ -123,15 +123,74 @@ where
 #[derive(Clone, Debug, Deserialize)]
 #[serde(untagged)]
 pub enum Actions {
+    // Allows keychords to map to null, which means no actions.
+    NoAction,
```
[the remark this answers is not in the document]
wistoft: To my knowledge it's needed, because an enum without a value, is what `serde` regards as a `null`.

It's only used for deserialization, and converted into `KeymapAction(vec![])` right away.

src/event_handler.rs
wistoft: Sure, I have rebased, and added a commit with an explanation.

k0kubun (approved):
:+1:

PR #550, opened 2024-11-01
Detect keyboards with mouse buttons by default

### LINKED ISSUE #483: Doesn't detect keyboards that support mouse functionality
First of all, great project! This really helped me getting my macos keyboard going with linux.

I think the assumption that a device can either only be a keyboard OR a mouse, but not both, like in: https://github.com/xremap/xremap/blob/f42ec7bdb4dbb2ec56baf3924aa592b4e9d57e2c/src/device.rs#L289-L301
is flawed though.

My keyboard for example is a Logitech MX Keys for Mac: https://www.logitech.com/en-us/products/keyboards/mx-keys-s-for-mac.920-011621.html?sp=2&searchclick=logi

And I guess because you could also buy the logitech MX mouse for it and connect them both to the same receiver, that's internally implemented as the mouse connecting to the keyboard and the keyboard showing up as both at the same time on the linux side. But even then, there's keyboards with integrated touchpads for example, which might show up as a keyboard and mouse too.

libinput also avoids putting devices into categories for this reason (because there's always some case where it doesn't work). https://who-t.blogspot.com/2015/06/libinput-and-lack-of-device-types.html

What's the problem with grabbing all devices by default? Is it for latency reasons? Maybe another way to check if a device needs to be grabbed would be to check if any of the remappings can be applied to the device (e.g. if the user chose to remap Z to Y, but the device doesn't have a KEY_Z, you don't need to grab the device)

NI57721:
fix #483

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -293,8 +293,6 @@ impl InputDevice {
                 keys.contains(Key::KEY_SPACE)
                 && keys.contains(Key::KEY_A)
                 && keys.contains(Key::KEY_Z)
-                // BTN_MOUSE
-                && !keys.contains(Key::BTN_LEFT)
             }
             None => false,
         }


k0kubun (approved):
👌 Let's try this. Sorry for not noticing the PR for a while for some reason.

NI57721:
Thank you very much!

PR #553, opened 2024-11-16
[wip] Support keyboard layout in config file

wistoft:
This PR is not finished, but is a proof of concept. For [issue](https://github.com/xremap/xremap/issues/552)

### Description

User will be able to define their keyboard layout in the config file, and then xremap respects that for all definitions in the config file.

### Sample config file

```yml
keyboard_layout:
  # Partial AZERTY layout
  q: a
  a: q

keymap:
  remap:
    control-a: k
```
Pressing the `control-a` on an AZERTY keyboard will trigger the defined remap. Even though evdev emits a `KEY_Q` to xremap.
 
### Implementation

Keyboard layout is applied before events are processed by xremap, and inversely applied when emitting keys. This means that xremap internally have some pseudo-keys, which the user can give a descriptive name.

The PR can be implemented as non-breaking, by letting users define a keyboard layout in the config file. If they don't have such configuration, the functionality would have no effect at all.

### Testing the PR

Change the hardcoded keyboard layout in [src/config/keyboard_layout.rs](https://github.com/xremap/xremap/pull/553/files#diff-91b4c370c7b1c37dfa4c3f85550665b2813939f05ef869c01d09fd63053aec0fR21). Then compile xremap as normal.

### DIFF
--- a/src/action_dispatcher.rs
+++ b/src/action_dispatcher.rs
@@ -8,6 +8,7 @@ use nix::sys::signal;
 use nix::sys::signal::{sigaction, SaFlags, SigAction, SigHandler, SigSet};
 use std::process::{exit, Command, Stdio};
 
+use crate::config::keyboard_layout::reverse_keyboard_layout;
 use crate::event::RelativeEvent;
 use crate::{action::Action, event::KeyEvent};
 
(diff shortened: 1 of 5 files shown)


theottm:
Great, I'll try that out!

Issue #492, opened 2024-06-12
Feature request: make `timeout_key` also accept a key sequence (currently it only accepts a single key combination)


hab25:
Also, `timeout_key` is not mentioned at all in the README; it is probably a good idea to document it thusly.

k0kubun:
Sounds good. Could you file the suggested change as a pull request? Thank you.

PR #565, opened 2025-01-01
Fix issue #492: timeout_key accept a key sequence

Yus314:
fix https://github.com/xremap/xremap/issues/492
# Feature
timeout_key  accept a key sequence :
```
keymap:
  - remap:
      J:
        remap:
          K: Esc
        timeout_key: [A, B, C]
        timeout_millis: 150
```

### DIFF
--- a/src/config/tests.rs
+++ b/src/config/tests.rs
@@ -163,7 +163,23 @@ fn test_yaml_keymap_remap() {
               C-s:
                 remap:
                   x: C-z
-            timeout_key: [Down]
+            timeout_key: Down
+            timeout_millis: 1000
+    "})
+}
+
+#[test]
+fn test_yaml_keymap_remap_timeout_as_sequence() {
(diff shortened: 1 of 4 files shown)


src/config/tests.rs
```diff
@@ -163,7 +163,7 @@ fn test_yaml_keymap_remap() {
               C-s:
                 remap:
                   x: C-z
-            timeout_key: Down
+            timeout_key: [Down]
```
k0kubun, line 166: It should be able to parse both a single key (without brackets) and a key sequence. Please fix the parser if it doesn't, and please leave tests for both cases if it does.

Yus314: Thank you for your review.

I have corrected the points you raised.
I have modified the parser to accept both single keys and key sequences and added test cases for both formats.

I would appreciate it if you could take a look when you have time.

Issue #571, opened 2025-01-05
Prevent double-type works, but not perfectly


berwynhoyt:
My spacebar frequently 'bounces' emitting two spaces. It's maddening. Perhaps a feature of how I type because it's now happened to me on two separate laptop keyboards. I'd love a way to prevent this without switching on bouncekeys, which creates other problems (cf. #563).

This solves the problem:
```
keymap:
  - name: prevent spacebar bounce from emitting two spaces
    remap:
      space:
        remap:
          space: null
        timeout_key: space
        timeout_millis: 150
```

This works already, and makes use of timeout_key (documented in #128) and recently merged `null` action (documented in #544). I will create a PR to document these soon.

Although the above does work, it isn't perfect because now the space always comes out 150ms after pressing the spacebar, which is noticable. Is it possible to do this instead:
```
keymap:
  - name: prevent spacebar bounce from emitting two spaces
    remap:
      space:
        press: space
        remap:
          space: null
        timeout_key: null
        timeout_millis: 150
```
Note the addition of `press: space` which I envisage will do the same as remap `space: space` except that it will also allow a timed nested remap to prevent bounce. Note also the change of timeout_key to `null` which prevents output of an additional space (since one was already put out immediately when space was pressed).

I'm asking two questions:
- Is there already a way that I'm unawareof to specify this `press: space` and still next the remap?
- If not, would someone consider adding this? Perhaps @wistoft or @kokubun? I do not know rust myself.

wistoft:
> Is there already a way that I'm unawareof to specify this press: space and still next the remap?

There isn't, it would be a new feature.

Issue #543, opened 2024-10-28
Document timeout_key


waffleabyssx0x:
`timeout_key` is not used in any of the examples, would be nice to have this option documented, either in the `README.md` or `example/config.yml`.

k0kubun:
Can you file a PR?

berwynhoyt:
> Can you file a PR?

@k0kubun This issue should be resolved once #572 is accepted.

waffleabyssx0x:
Thank you! Closing this issue.

Issue #128, opened 2022-07-28
Support timeout_key on nested remap


kohane27:
Hello @k0kubun. Hope you're doing well. Thank you for creating this awesome remap software. This is the only tool I find working reliably under Wayland and inside Wine!

However, I have trouble figuring out the config syntax. What I hope to acheive: `jk` is remapped to `KEY_ESC`; for a single `j` remapped to `KEY_DOWN`

I tried the following but it doesn't work:

```
keymap:
  - name: balabolka
    application:
      only: [balabolka.exe]
    remap:
      KEY_J:
        remap:
          KEY_K: KEY_ESC
          KEY_J: KEY_DOWN
```

The following works:

```
      KEY_J:
        remap:
          KEY_K:
            KEY_ESC
        timeout_millis: 150 # optional
```

But a single `KEY_J` will output the literal `j`, which is not what I want. I want a single `KEY_J` to output `KEY_DOWN`.

Any help is much appreciated. Thank you!

k0kubun:
It seems like your use case is currently not implemented. Honestly, I haven't even thought of that.

To support that, a new syntax like this would be needed:

```yml
keymap:
  - name: balabolka
    application:
      only: [balabolka.exe]
    remap:
      KEY_J:
        remap:
          KEY_K:
            KEY_ESC
        timeout_key: KEY_DOWN # new
        timeout_millis: 150
```

As I'm currently focusing on a big change for another OSS, I'm planning to work on this issue on Aug 13th. If I receive a pull request before that, I can take a look at that too.

kohane27:
Hello @k0kubun , thank you for getting back to me on such short notice; I appreciate it:)

Thank you for considering implementing this feature. Honestly the only use case I have for such two-key remapping is making a primitive Vim-like modal editing. The only use case I have for this is just "escaping to normal mode" with `jk`. I'm not sure how many people need this feature and it's completely up to you (as always) to see if it's worth implementing. I have zero knowledge with Rust; my apologies.

I looked into [PR#93](https://github.com/k0kubun/xremap/pull/93) but it seems such two-key remapping is not possible through vim mode.

I look forward to hearing from you:)

k0kubun:
Released https://github.com/k0kubun/xremap/pull/144 in v0.5.2.

PR #572, opened 2025-01-05
Document timeout_key and null actions in README

berwynhoyt:
- Documents the use of `timeout_key` from #128 (using a use-case from #571).
  - This also resolves outstanding issue #543 for @waffleabyssx0x as requested by @k0kubun.
- Also documents `null` keymap actions per #544.


k0kubun (approved):
:+1:

PR #573, opened 2025-01-05
Clarify README to indicate that the KEY_XXX example keys are all unique, not the same key

berwynhoyt:
Previously it wasn't clear that the example expected different keys in each of these places, and could confuse the reader.


k0kubun (approved):
👌

PR #574, opened 2025-01-05
Add sleep KeymapAction

### LINKED ISSUE #535: Feature: Delay within sequence
Currently it seems the only way to define a delay is globally, using `keypress_delay_ms`. 

It would be nice to also be able to define delays within key sequences, such as:

```yaml
      CapsLock-n: [
          Shift-left, 
          Ctrl-c, 
          launch: ["/bin/python", "/home/ascyt/.config/xremap/script/format_clipboard.py"],
          # I would like to have a delay here, or to be able to wait until the above script has finished executing. 
          Ctrl-v
        ]
```
The above code basically, when in a text field, copies the last character to clipboard, runs a python script (which changes the clipboard), then pastes the result. I figure the reason why this doesn't work (when I do Ctrl-v manually, it does paste the correct character) is because the Python script has not had time to finish executing before xremap did Ctrl-v.  

I haven't found a way in the README.md to do this. It would be nice to, for example, put `delay_ms:` within a sequence, and/or something like `launch-wait:` to launch a program, but wait for it to finish executing before continuing the sequence.

wistoft:
Fix #535

Adds a new enum: `KeymapAction::Sleep`, which can be used to sleep when emitting actions.

The sleep action can be used for press/release keys and in keymap:

```yml
modmap:
  - remap:
      # When releasing f12 it will sleep 500ms before emitting b.
      f12:
        press: a
        release: [{ sleep: 500 }, b]

keymap:
  - remap:
      # Will emit KEY_A rightaway, sleep 500ms and then emit KEY_B.
      f11: [a, { sleep: 500 }, b]
``` 

The sleep is synchronous and will block all processing, so key events will queue up. It will for instance not be possible to stop xremap by typing `control-c` in the terminal, until the sleep is done.

If the mouse is captured, then the mouse will be unresponsive while sleeping. It can't move as all its events is queue up.

### DIFF
--- a/src/config/keymap_action.rs
+++ b/src/config/keymap_action.rs
@@ -29,6 +29,8 @@ pub enum KeymapAction {
     WithMark(KeyPress),
     #[serde(deserialize_with = "deserialize_escape_next_key")]
     EscapeNextKey(bool),
+    #[serde(deserialize_with = "deserialize_sleep")]
+    Sleep(u64),
 
     // Internals
     #[serde(skip)]
@@ -117,6 +119,19 @@ where
         }
     }
     Err(de::Error::custom("not a map with a single \"escape_next_key\" key"))
+}
+
+fn deserialize_sleep<'de, D>(deserializer: D) -> Result<u64, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let mut action = HashMap::<String, u64>::deserialize(deserializer)?;
+    if let Some(set) = action.remove("sleep") {
+        if action.is_empty() {
+            return Ok(set);
+        }
+    }
+    Err(de::Error::custom("not a map with a single \"sleep\" key"))
 }
 
 // Used only for deserializing Vec<Action>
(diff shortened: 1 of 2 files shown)


berwynhoyt:
I can confirm that this PR works well. Please go ahead and merge it. It can be used for various purposes, including making certain key mappings work, even with bouncekeys.

For the sake of completeness, below is an example that needs it to work with bouncekeys. The example remaps `Ctrl-delete` to delete the subsequent word _including trailing space_ because, whereas Ctrl-delete already deletes trailing space in many apps, it does not, by default, in Chrome, Firefox, and various web-based apps:
```yml
shared:
  sleep: &sleep {sleep: 0}

keymap:
  - name: del_next_word_and_trailing_spaces
    remap:
      C-delete: [{set_mark: false}, C-shift-right, *sleep, C-shift-right, *sleep, C-shift-left, delete]
```
In the above, the sleep is a workaround for if you use bouncekeys. You will need to increase the sleep time in ms to be larger than your bouncekeys delay (in Linux Mint, bouncekeys is in the accessibility/typing settings and each slider 'notch' is 10ms). This is because these key sequences output repetitions of the same keystroke (most notably Ctrl and Shift) which bouncekeys suppresses. Note that the problem only occurs when bouncekeys is enabled after xremap is run.

k0kubun (approved):
:+1:

PR #575, opened 2025-01-07
Allow matching any key

apatard:
Current configuration is allowing to match a key and remap it. While it is useful for must cases, it is less useful when trying to disable shortcuts. If there are a lot of shortcuts to disable, it may be easier (and better for security) to disable all shortcuts except some.

This PR is adding a new "virtual" key used to match all keys. For instance, using:
```
keymap:
- remap:
    a: b
    ANY: null
```
will result in disabling all keyboards keys except ``KEY_A``, which is remapped to ``KEY_B``.

In case of shortcuts, this will disable all ``C-$KEY`` except ``C-p``:
```
keymap:
- exact_match: true
  remap:
    C-p: C-p
    C-ANY: reserved
```

### DIFF
--- a/src/config/key.rs
+++ b/src/config/key.rs
@@ -1,4 +1,4 @@
-use crate::event_handler::DISGUISED_EVENT_OFFSETTER;
+use crate::event_handler::{DISGUISED_EVENT_OFFSETTER, KEY_MATCH_ANY};
 use evdev::Key;
 use serde::{Deserialize, Deserializer};
 use std::error::Error;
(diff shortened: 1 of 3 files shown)


src/config/key.rs
```diff
@@ -119,6 +119,7 @@ pub fn parse_key(input: &str) -> Result<Key, Box<dyn Error>> {
             REL_WHEEL_HI_RES = 0x0b,
             REL_HWHEEL_HI_RES = 0x0c,
         */
+        "ANY" => Key(DISGUISED_EVENT_OFFSETTER + 26),
```
k0kubun, line 122: 26 feels like a magic number at a glance. Could you make a constant that hides this implementation detail, add a clarifying comment above the constant, and use the constant here?

apatard: I've updated the PR to use a constant. I hope I understood correctly what you meant.

k0kubun (approved):
👌

PR #577, opened 2025-01-28
Ignore absolute input devices

apatard:
Xremap is reporting itself as an input device sending relative events. When it's started with ``--mouse``, it will also grab input devices sending absolute events and thus is breaking them.

Until xremap is supporting absolute devices, simply ignore them for now.


src/device.rs
```diff
@@ -301,6 +302,11 @@ impl InputDevice {
     }
 
     fn is_mouse(&self) -> bool {
+        // Xremap doesn't support absolute device so will break them.
+        if self.device.supported_absolute_axes().is_some()  {
```
k0kubun, line 306: Where did you define this `supported_absolute_axes()`?

k0kubun (approved):

PR #582, opened 2025-02-09
Fix static_mut_refs warning

wistoft:
Rust warns about static_mut_refs in device.rs. Info: [https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html](https://doc.rust-lang.org/nightly/edition-guide/rust-2024/static-mut-references.html)

This PR fixes this warning by creating a thread local variable for `DEVICE_NAME`. This is possible because `current_name` is ever only called from the main thread.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -6,7 +6,6 @@ use derive_where::derive_where;
 use evdev::uinput::{VirtualDevice, VirtualDeviceBuilder};
 use evdev::{AttributeSet, BusType, Device, FetchEventsSynced, InputId, Key, RelativeAxisType};
 use nix::sys::inotify::{AddWatchFlags, InitFlags, Inotify};
-use std::cell::RefCell;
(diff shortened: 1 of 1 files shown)


k0kubun:
Can you fix this without using a thread local? Using a thread local variable is often a slight performance hit for a single-threaded application, and xremap is single-threaded. Using a thread local doesn't seem like the right way to fix the warning to me. I don't think we take multiple references to this `static mut` at once either.

For example, you should be able to suppress this warning by `#[allow(static_mut_refs)]`. Since it's most likely not going to be UB for this `static mut`, it seems like a false positive. There's nothing wrong in suppressing a false-positive warning.

k0kubun:
I took the liberty to fix it that way. Merging.

wistoft:
I agree, it's also fine to ignore it.

PR #583, opened 2025-02-13
Allow specifying uinput device product / vendor IDs

apatard:
Current code is hard-coding the IDs of the uinput device. Add code to keep it as default and to specify product and/or vendor IDs.


k0kubun:
The change failed the CI. Could you fix the test?

PR #584, opened 2025-02-13
Allow matching a device on input device vendor/product IDs

apatard:
Instead of matching on a device only its name or path, allow using its vendor/product IDs too.


k0kubun:
This PR also seems to fail the CI. Could you fix the test?

k0kubun (approved):

PR #587, opened 2025-02-19
Allowing matching event device on udev properties

apatard:
This is a follow-up of PR584. When the device IDs are not known, it's still
possible to match them if they are sharing a common udev property (as shown with
``udevadm info /dev/input/eventX``).

So, add the udev crate and use it to get the device property to match.
Given that it may not be desirable for everyone, this has been put
inside a feature. Moreover, this will avoid build issues in the CI,
due to the musl toolchain missing the udev development files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,30 @@ jobs:
           name: xremap-${{ matrix.arch }}-${{ matrix.feature }}
           path: target/${{ matrix.arch }}-unknown-linux-musl/release/xremap-linux-${{ matrix.arch }}-${{ matrix.feature }}.zip
 
+  build-libc:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        feature: [udev]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+          target: x86_64-unknown-linux-gnu
+      - uses: Swatinem/rust-cache@v2
+        with:
+          key: ubuntu-latest-glibc-${{ matrix.feature }}
+
+      - name: Install libudev dependencies
+        run: sudo apt-get update && sudo apt-get -y install pkg-config libudev1 libudev-dev
+
+      - name: Build
+        run: cargo build --release --features ${{ matrix.feature }}
+
   test:
     runs-on: ubuntu-latest
     steps:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -850,6 +850,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "libudev-sys"
+version = "0.1.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c8469b4a23b962c1396b9b451dda50ef5b283e8dd309d69033475fa9b334324"
+dependencies = [
+ "libc",
+ "pkg-config",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.3.8"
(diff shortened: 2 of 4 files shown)


k0kubun:
> Moreover, this will avoid build issues in the CI, due to the musl toolchain missing the udev development files.

Our single-binary distribution relies on musl compatibility. If xremap has this feature enabled, we can't release such xremap in binary releases. I don't want to have a separate cargo feature that is not going to be released.

How about maybe invoking a command like `udevadm` instead of using API that's not implemented in musl?

apatard:
> > Moreover, this will avoid build issues in the CI, due to the musl toolchain missing the udev development files.
> 
> Our single-binary distribution relies on musl compatibility. If xremap has this feature enabled, we can't release such xremap in binary releases. I don't want to have a separate cargo feature that is not going to be released.
> 
> How about maybe invoking a command like `udevadm` instead of using API that's not implemented in musl?

There are several reasons  I used the crate:
- it's doing all the work of handling udev. Nothing to parse
- I'm not sure the udevadm output is stable
- calling ``udevadm`` will be slower.

k0kubun:
I understood you want to use the crate. If you want to merge this, can you fix the following things?

1. There should be at least one GitHub Actions workflow that tests this code.
    * Because you marked the feature as optional and you didn't modify any workflow, none of the existing CI workflows are testing your code.
2. If you want this feature in the binary distribution of the xremap repository, update the GitHub Actions release workflows as such.
    * If you're fine with building your xremap using `crate` command manually to enable this feature, you may skip this step.
    * If you decide to update the release workflow, you still have to use musl-libc and the workflows should run successfully.
3. Cargo.toml has a conflict. You need to rebase your branch from master.

apatard:
> I understood you want to use the crate. If you want to merge this, can you fix the following things?
> 
>     1. There should be at least one GitHub Actions workflow that tests this code.
>        
>        * Because you marked the feature as optional and you didn't modify any workflow, none of the existing CI workflows are testing your code.

oh, sure. I wanted to be sure of the direction to go before doing anything more.

> 
>     2. If you want this feature in the binary distribution of the xremap repository, update the GitHub Actions release workflows as such.
>        
>        * If you're fine with building your xremap using `crate` command manually to enable this feature, you may skip this step.

I'm totally fine building xremap manually and I guess it's best to do it this way. There will always be someone not being able to run the binary from release due to different libc/udev/whatever/... 

>        * If you decide to update the release workflow, you still have to use musl-libc and the workflows should run successfully.
> 
>     3. Cargo.toml has a conflict. You need to rebase your branch from master.

Obviously. As mentioned for point 1, I wanted to get some feedback before doing anything.

Thanks

Cross.toml
```diff
@@ -0,0 +1,8 @@
+[target.x86_64-unknown-linux-gnu]
+pre-build = [
+    "apt-get update && apt-get -y install pkg-config libudev1 libudev-dev"
+]
+[target.aarch64-unknown-linux-gnu]
+pre-build = [
+    "dpkg --add-architecture arm64 && apt-get update && apt-get -y install pkg-config libudev1:arm64 libudev-dev:arm64"
+]
```
k0kubun, line 8: This seems very specific to platforms using apt-get/dpkg. I feel like this should be done on the definition of the GitHub Actions workflow, not here.

Also, because we're not going to ship this in binary releases, you don't need to support `cargo cross` in the first place. Just run `cargo build --release` in the workflow please.

apatard: ok. I re-used what was done for the other step, so I had to cope with the way ``cross`` was working. I've updated to build again the system without cross. I've also removed the aarch64 build, to keep things simple.

k0kubun (approved):

apatard:
oh, great. Thanks for merging this PR

PR #643, opened 2025-02-24
feat: enable LTO and codegen-units = 1 optimizations

zamazan4ik:
Resolves https://github.com/xremap/xremap/discussions/538


k0kubun (approved):
:+1: Thanks!

PR #665, opened 2025-04-10
Uses the correct parser for secondary toml files

frflcn:
Uses the correct parser for 2nd, 3rd, 4th ... etc .toml files


k0kubun (approved):
:+1:

PR #673, opened 2025-04-17
Re-add AddWatchFlags if config file has been deleted then recreated

frflcn:
## What's been changed

Re-adds AddWatchFlags if config file has been deleted then recreated or has been overwritten by renaming another file to its own name.

## Why

If a file has been deleted and recreated or has been overwritten by renaming another file to its own name. The watch flags specific to the file (not the WatchFlags on its parent directory) will be dropped and certain types of modifications to the file will not be notified. 

This is particularly troublesome if the user makes some changes in an editor such as [xed](https://github.com/linuxmint/xed) and saves, then switches to an editor such as VSCode, makes some more changes and then saves. The changes in VSCode will not be picked up and any further changes will not be picked up. Changes in xed however will continue to work. Changes in VSCode would have worked prior to saving with xed.

This is because when xed saves a file, it first creates a temporary file then renames it to the file the user expects. The operating system sees this renaming operation as a deletion of the old file and so drops our old WatchFlags on the file. Any operations that would normally trigger those flags will no longer work, which is why VSCode(who triggers on the file) stops triggering changes, but xed(who triggers flags on its directory) still triggers.

## How to reproduce the errors 

Running xremap with `--watch=config`

### Method 1:

1. Delete the config file. (No change occurs to the mapping _(fine if intended)_)
2. Recreate the file with VSCode. _**(xremap stops working)**_

### Method 2:

1. Edit the file with VSCode and save (Changes take effect)
3. Edit the file with xed and save (Changes take effect)
4. Edit the file with VSCode and save _**(No changes take effect)**_
5. Edit the file with xed and save (Changes take effect)

## Other info

This isn't particular to VSCode and xed but to any editor who saves files in a similiar manner

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -174,6 +174,7 @@ fn main() -> anyhow::Result<()> {
                         &ignore_filter,
                         mouse,
                         &config_paths,
+                        inotify,
                     )? {
                         break 'event_loop ReloadEvent::ReloadConfig;
                     }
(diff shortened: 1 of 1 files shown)


frflcn:
Let me know if deleting a config file having no effect on the mapping is intended or not. If not I can create a pull request for it as well.

k0kubun (approved):
👌

PR #678, opened 2025-04-30
Add mapping for press and release key

frflcn:
[Someone was asking](https://github.com/xremap/xremap/discussions/618#discussioncomment-12964476) about a scenario that seemed to need switching between modes upon release of a key. I directed them towards PressReleaseKey of modmap. It seemed to solve their issue, but then they said they also wanted to map the key to a different one. This isn't supported by PressReleaseKey nor could I find anything to solve their issue in xremap. So I decided to add it to PressReleaseKey.

### This pull request:

- Adds an optional field, called "remap" to PressReleaseKey that takes a Key which the PressReleaseKey will be mapped to when set. When remap is set the skip_key_event field on PressReleaseKey has no effect (because I'm assuming the user will naturally desire this). If you would like it otherwise I can change it.
- Updates the README to include the remap field AND makes it more clear that while press and release are required they can be null if they want an event fired on only one of them.
- Adds a test for the new mapping for PressReleaseKey

#### With remapping:
```yaml
modmap:
 - remap:
     KEY_A:
       remap: KEY_B
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

#### Without mapping:
```yaml
modmap:
 - remap:
     KEY_A:
       press: { set_mode: "on" }
       release: { launch: "bash" }
```
##### OR
```yaml
modmap:
 - remap:
     KEY_A:
       remap:
       press: { set_mode: "on" }
       release: { launch: "bash" }
```
##### OR
```yaml
modmap:
 - remap:
     KEY_A:
       remap: null
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

#### When remap is set
```yaml
modmap:
 - remap:
     KEY_A:
       skip_key_event: true # has no effect was going to skip the key event anyways
       remap: KEY_B
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

```yaml
modmap:
 - remap:
     KEY_A:
       skip_key_event: false # HAS NO EFFECT (will still skip the original key event) If this isn't desirable I can change it
       remap: KEY_B
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

#### When remap is not set, skip_key_event behaves normally

```yaml
modmap:
 - remap:
     KEY_A:
       skip_key_event: false # has effect will not skip the original key event
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

```yaml
modmap:
 - remap:
     KEY_A:
       skip_key_event: true # has effect will skip the original key event
       press: { set_mode: "on" }
       release: { launch: "bash" }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,8 +207,9 @@ modmap:
       # Hook `keymap` action on key press/release events.
       KEY_XXX3:
         skip_key_event: false # Optional, skip original key event, defaults to false
-        press: { launch: ["xdotool", "mousemove", "0", "7200"] } # Required
-        release: { launch: ["xdotool", "mousemove", "0", "0"] } # Required
+        remap: KEY_YYYY # Optional, replace the key with another. If set skip_key_event does nothing
+        press: { launch: ["xdotool", "mousemove", "0", "7200"] } # Required, but can be null
+        release: { launch: ["xdotool", "mousemove", "0", "0"] } # Required, but can be null
     application: # Optional
       not: [Application, ...]
       # or
--- a/README.md
+++ b/README.md
@@ -207,9 +207,8 @@ modmap:
       # Hook `keymap` action on key press/release events.
       KEY_XXX3:
         skip_key_event: false # Optional, skip original key event, defaults to false
-        remap: KEY_YYYY # Optional, replace the key with another. If set skip_key_event does nothing
-        press: { launch: ["xdotool", "mousemove", "0", "7200"] } # Required, but can be null
-        release: { launch: ["xdotool", "mousemove", "0", "0"] } # Required, but can be null
+        press: { launch: ["xdotool", "mousemove", "0", "7200"] } # Required
+        release: { launch: ["xdotool", "mousemove", "0", "0"] } # Required
     application: # Optional
       not: [Application, ...]
       # or
--- a/README.md
+++ b/README.md
@@ -206,9 +206,10 @@ modmap:
         alone_timeout_millis: 1000 # Optional
       # Hook `keymap` action on key press/release events.
       KEY_XXX3:
-        skip_key_event: false # Optional, skip original key event, defaults to false
-        press: { launch: ["xdotool", "mousemove", "0", "7200"] } # Required
-        release: { launch: ["xdotool", "mousemove", "0", "0"] } # Required
+        skip_key_event: true # Optional, skip original key event, defaults to false
+        press: [{ press: KEY_YYY }, { launch: ["xdotool", "mousemove", "0", "7200"] }] # Optional
+        repeat: { repeat: KEY_YYY } # Optional
+        release: [{ release: KEY_YYY }, { set_mode: my_mode }] # Optional
     application: # Optional
       not: [Application, ...]
       # or
--- a/src/config/key.rs
+++ b/src/config/key.rs
@@ -4,20 +4,6 @@ use serde::{Deserialize, Deserializer};
 use std::error::Error;
 use std::str::FromStr;
 
-pub fn deserialize_optional_key<'de, D>(deserializer: D) -> Result<Option<Key>, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    let option_key_string: Option<String>  =  Option::deserialize(deserializer)?;
-    if let Some(key_string) = option_key_string {
-        let key = parse_key(&key_string).map_err(serde::de::Error::custom)?;
-        return Ok(Some(key));
-    }
-    else {
-        return Ok(None);
-    }
-}
-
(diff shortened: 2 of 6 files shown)


k0kubun:
The interface seems overly specific to the discussed use case. Could you consider changing it to either of the following interfaces (or support both)?

```yml
modmap:
 - remap:
     KEY_A:
       press: [KEY_B, { set_mode: "on" }] # press B and set mode on
       release: [KEY_B, { launch: "bash" }] # release B and launch bash
```

```yml
modmap:
 - remap:
     KEY_A:
       press: [{ press: KEY_B }, { set_mode: "on" }] # press B and set mode on
       release: [{ release: KEY_B }, { launch: "bash" }] # release B and launch bash
```

frflcn:
Sure! This seems like it would be a great thing to add. A couple of things I would like to talk about first.

### Adding  a repeat field

Since we are dropping the remap field I had added, I think it would be a good idea to add an optional repeat field. That way the user can have a true modmap remapping on the press and release key via something like this:

```yaml
modmap:
 - remap:
     KEY_A:
       skip_key_event: true
       press: [{ press: KEY_B }, { set_mode: "on"}]
       repeat: [{ repeat: KEY_B }]
       release: [{ release: KEY_B }, { launch: "bash" }]
```

### The different options you mentioned

I can definitely do the second option you mentioned. As for the first option, you can actually write a yaml like that right now but it has different functionality then what you are asking for. The current functionality is like this:
```yaml
modmap:
 - remap:
     KEY_A:
       press: [KEY_B, { set_mode: "on" }] # press B, release B and set mode on
       release: [KEY_B, { launch: "bash" }] # press B, release B and launch bash
```
This stems from the fact that press and release actually takes a Vec\<KeymapAction\>. So KEY_B gets deserialized as a KeymapAction::KeyPress, which takes one or more keys separated by a dash '-' and gets dispatched as a keychord.

### Make the press and release fields optional?

While we are talking about press and release key. The README states that press and release are required, but if the user wants to only fire an event on one of them, it doesn't demonstrate that it can be empty. Like this:
```yaml
modmap:
 - remap:
     KEY_A:
       press: { set_mode: "on }
       release:
```
It does demonstrate it in the section about modes, but that's not where you'd go to understand something like this.

I feel like we should either add an explicit demonstration in the README or make press and release optional. If we make press and release optional then we could write something like this that achieves the same thing:
```yaml
modmap:
 - remap:
     KEY_A:
       press: { set_mode: "on }
```

I feel like this is a little bit more intuitive way to write it. What do you think?

### Rename KeymapAction::KeyPress to KeymapAction::KeyChordPress

One final thing I'd like to mention is: To make room for { press: KEY_A } I think I'll have to rename KeymapAction::KeyPress (that I mentioned above) to a more descriptive name like KeymapAction::KeyChordPress, then { press: KEY_A } will take the name KeymapAction::KeyPress. I feel like this makes the most intuitive sense, but I wanted to get your approval before I go ahead making changes to type names.
<br><br>
I know I've written alot here. Let me know what you think.

k0kubun:
> As for the first option, you can actually write a yaml like that right now but it has different functionality then what you are asking for. The current functionality is like this

OK, that's fine. For now, let's keep the backward compatibility.

> Since we are dropping the remap field I had added, I think it would be a good idea to add an optional repeat field

:+1: It sounds good to add a `repeat` field.

> I feel like we should either add an explicit demonstration in the README or make press and release optional.

Let's make it optional if you want that feature.

> One final thing I'd like to mention is: To make room for { press: KEY_A } I think I'll have to rename KeymapAction::KeyPress (that I mentioned above) to a more descriptive name like KeymapAction::KeyChord, then { press: KEY_A } will take the name KeymapAction::KeyPress

I'm not opposed to adding the `KeyChord` feature later, but for now, let's keep the backward compatibility of just writing `KEY_A`. You may rename it to `KeymapAction::KeyPressAndRelease` for clarity. Then you should be able to use `KeymapAction::KeyPress` for `{ press: KEY_A }`.

frflcn:
Ok I've made the changes we talked about. Only a couple of things to note.


### Renaming
 - I renamed KeymapAction::KeyPress to KeymapAction::KeyPressAndRelease like we talked about. 
 - I also renamed a couple of functions that referenced KeymapAction::KeyPress to reflect the changes. 
 - KeymapAction::KeyPress(as it was called) takes a struct that is also called KeyPress, but I did not rename the struct because it is used elsewhere in the codebase to fire a Keymap and in that instance it fires on press, and release has very little relevance there, so it didn't make sense to change the structs name to KeyPressAndRelease as well.

### README
 - I updated the README to reflect the changes in PressReleaseKey

k0kubun (approved):
:+1:

PR #705, opened 2025-07-09
Code cleanups, refactors, bug fixes, optimizations and new `free_hold` feature

prime-run:
This PR:

Adds new feature: 

   - `config`: Add `free_hold` option for multi-purpose keys
     - A new boolean option, free_hold, has been added to the multi-purpose key configuration.
     - When free_hold: true, the held action is only triggered after another key is pressed, rather than after a timeout. This allows a key to be held
       indefinitely without triggering its "held" state, which is ideal for keys that also serve as modifiers.

Fixes:

   - `event-handler`: Prevent duplicate key presses when held by reordering dispatchers
     - The event handler now correctly filters out redundant key presses that are already part of a flushed event queue.

Refactors:
   - `event-handler` : Improve `MultiPurposeKeyState` logic
     - The internal state management for multi-purpose keys has been significantly refactored for better readability and robustness. Complex conditional
       logic has been simplified into more straightforward, pattern-matching-based structures.

   - `event-handler` : Streamline `dispatch_keys` method
     - The `dispatch_keys` method in the event handler has been refactored to be more concise and efficient, improving the overall clarity of the
       event-handling pipeline.

Optimizes: 

   - `event-handler`: Use ***slices for more idiomatic arguments***
     - `Vec<T>` -> `&[T]` for fn args. This avoids unnecessary allocations and provides more flexibility to the caller.

   - Minor code cleanup
     - Removed a redundant conditional check against `true` , simplifying the logic for handling relative mouse events. (Update: reverted due to failing tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,9 +277,27 @@ sudo RUST_LOG=debug xremap config.yml
 
 Then press the key you want to know the name of.
 
-If you specify a map containing `held` and `alone`, you can use the key for two purposes.
-The key is considered `alone` if it's pressed and released within `alone_timeout_millis` (default: 1000)
-before any other key is pressed. Otherwise it's considered `held`.
+If you specify a map containing `held` and `alone`, you can use the key for two purposes. By default, the behavior is determined by a timeout:
+- If the key is pressed and released within `alone_timeout_millis` (default: 1000) without any other key being pressed, it's considered `alone`.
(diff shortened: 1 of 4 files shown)


k0kubun:
I'll accept this feature. Could you update README to document it? It's a bit complicated switch, so it seems helpful to have an explanation.

PR #718, opened 2025-08-05
Implement Held Threshold and rename Alone to Tap

Tradgardsmastaren:
## Summary

This PR attempts to implement the behavior proposed in #614, extending the tap/hold options for more flexible multi-purpose key handling.

## New Behavior

A key now transitions through distinct states based on timing and whether another key is pressed:

**Key behavior:**
- **Before `held_threshold_millis`:**  
  - If released alone → **tap**  
  - If another key is pressed → **tap**
  
- **After `held_threshold_millis` and before `tap_timeout_millis`:**
  - If another key is pressed → **held**
  - If released alone → **tap**

- **After `tap_timeout_millis`:**  
  - Always **held**

The default held_threshold_millis is set to `0`.
## Potential Bug Fix

This also appears to fix a bug that exists in the current `master` branch but was not present in version `0.10.11` (which I had previously installed). The issue occurs with modifier-tap keys (e.g. tap = `SHIFT_L`, hold = `CTRL_L`):

Typing quickly (e.g. "SHIFT, S") could cause the key to send `CTRL_L + S`, but incorrectly leave `CTRL_L` held even after release, since the release event was processed as `SHIFT_L`.

## Notes

- I adopted the naming (`held`, `held_threshold_millis`, `tap`, `tap_timeout_millis`) based on the discussion. This may break backward compatibility, and I can rename it if needed.
- I’ve tested a few configurations and they behaved as I expected, but I haven’t written formal tests. If tests are required, I’d be happy to look into it and would appreciate some guidance on how to proceed.
- I'm fairly new to open source development, so if I’ve done anything incorrectly (in code, structure, or process) I apologize and welcome any feedback or corrections.
- Let me know if anything else should be changed.

Best regards,  
**Trädgårdsmästaren**

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 - Remap any keys, e.g. Ctrl or CapsLock.
 - Remap any key combination to another, even to a key sequence.
 - Remap a key sequence as well. You could do something like Emacs's `C-x C-c`.
-- Remap a key to two different keys depending on whether it's pressed alone or held.
+- Remap a key to two different keys depending on whether it's tapped or held.
 - Application-specific remapping. Even if it's not supported by your application, xremap can.
 - Device-specific remapping.
 - Automatically remap newly connected devices by starting xremap with `--watch`.
@@ -234,11 +234,11 @@ modmap:
     remap: # Required
       # Replace a key with another
       KEY_XXX1: KEY_YYY # Required
-      # Dispatch different keys depending on whether you hold it or press it alone
+      # Dispatch different keys depending on whether you hold it or tap it
       KEY_XXX2:
         held: KEY_YYY # Required, also accepts arrays
-        alone: KEY_ZZZ # Required, also accepts arrays
-        alone_timeout_millis: 1000 # Optional
+        tap: KEY_ZZZ # Required, also accepts arrays
+        tap_timeout_millis: 1000 # Optional
       # Hook `keymap` action on key press/release events.
       KEY_XXX3:
         skip_key_event: true # Optional, skip original key event, defaults to false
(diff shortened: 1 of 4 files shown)


k0kubun:
To somewhat ease the migration to the new key names, is it possible for you to take a look at also supporting `held`/`alone`/`alone_timeout` as an alternative way to specify `hold`/`tap`/`tap_timeout`? I'll drop that in some later version, but it'd be nice to have a migration period if possible.

Tradgardsmastaren:
It should support both `alone/held` and `tap/hold` now, but I don't believe it supports both `alone` and `tap` in the same config if that is necessary. (and the same for `hold` and `held`)

While fixing your comment I noticed some behaviour which wasn't intended. If I tested with this config
```yml
modmap:
  - name: test
    remap:
      q:
        hold: Shift_L
        hold_threshold_millis: 200
        tap: q
        tap_timeout_millis: 1000
      w:
        hold: Ctrl_L
        hold_threshold_millis: 200
        tap: w
        tap_timeout_millis: 1000
      e:
        hold: Alt_L
        hold_threshold_millis: 200
        tap: e
        tap_timeout_millis: 1000
```
and wrote `qw` rapidly it would write wq since they both got flushed on the release of the `q`-key.
Since the `flush_timeout_keys` function puts the flushed keys first it would emit the w first. But simply putting them in the other way around wouldn't work either because if it was a normal key that was pressed you'd want the flushed key first.

Thus I added a workaround so that if a multikey is released it only flushes the multikeys that were pressed before it.
Now the code behaves as I envisioned the requested changes should. But I'm not sure how good it is, both in terms of code quality and performance since I am fairly new to Rust. So please have a look and see if it is up to the standards of the project or not.

// Trädgårdsmästaren

k0kubun:
looks good.

> I added a workaround so that if a multikey is released it only flushes the multikeys that were pressed before it.

Could you consider adding a test case for this in `src/tests.rs`, using `assert_actions`? If you can't reasonably do this due to the time-dependent behavior, you can say so. Using a very long threshold (or 0) might be a quick workaround without using a stub for the clock.

ruanswanepoel:
Any update on this PR? I just installed xremap and the lack of a held threshold is the biggest issue I have right now.

k0kubun:
Sorry, I didn't realize there was a new commit pushed to this PR.

@Tradgardsmastaren Could you fix the conflict please? And once you're done, it'd be nice if you can leave a comment to indicate it's fixed so that I get notified to merge this.

k0kubun:
@ruanswanepoel Could you test if this satisfies your needs? If your test goes well, I'll ship it.

ruanswanepoel:
I did find a weird bug that will need some explaining, I'm not sure if it's related to this PR.

I have this in my config:
```yml
K:
  tap: K
  held: SHIFT_R
  held_threshold_millis: 200
```

If I hold 'k' and press for example 'j' in less than 200ms, my output is 'kj' as expected. But if I continue to hold 'k' without releasing it, and then I press 'j' again (after 200ms since I started pressing 'k'), then it will output a capital 'J'. So it is registering that I'm hold 'k' down even after the fact that it's already registered a 'k' tap and 'j' tap event. That's actually not a bug but a feature IMO, but sadly there's more. Now after all this if I release 'k' you'd expect everything to go back to normal, but it **mistakenly still thinks I'm holding 'k' in this state**. I have to tap 'k' again to "release" it here. 

I hope that makes sense.

PR #720, opened 2025-08-11
Add Niri client

saurabhsharan:
Add a client for the [Niri](https://github.com/YaLTeR/niri) compositor to support application/window-specific remappings.

It should work as expected just like all the other WM/DE-specific plug-ins - I was able to re-use my config.yml that I used on KDE exactly as-is with Niri. I also confirmed that the `application:` and `window:` matchers both work as expected.

Notes:
- Even though I added a release matrix entry for `(aarch64, niri)`, I have only tested this on x86 as I don't have an ARM Linux machine.
- Niri [supports X11 windows](https://github.com/YaLTeR/niri/wiki/Xwayland) via xwayland-satellite, but I have not tested this with any X11 windows in Niri.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: [x86_64, aarch64]
-        feature: [x11, gnome, kde, hypr, wlroots]
+        feature: [x11, gnome, kde, hypr, wlroots, niri]
     steps:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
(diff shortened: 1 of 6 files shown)


k0kubun:
Thanks for the PR.

I'm open to merging this feature, but could you be a maintainer for the Niri support after that? I will not be able to triage/fix issues for the Niri integration myself.

saurabhsharan:
> Thanks for the PR.
> 
> I'm open to merging this feature, but could you be a maintainer for the Niri support after that? I will not be able to triage/fix issues for the Niri integration myself.

Yup, that's fine with me.

k0kubun (approved):

PR #731, opened 2025-09-02
Fix free_hold bug, add full coverage event dispatching tests, and refactor event_handler.

prime-run:
This PR fixes the bug inroduced in #724 causing relesse event fail to dispatch when event chain concluded before definec `alone_timeout_millis`.

And also add missing test scenarios to include all possible and realistic scanrios regarding events by testing against `alone_timeout_millis` timeout window with and without `free_hold`.

And also Refactors `event_handler.rs` file by breaking it into smaller modules + genreal refactors to maintain readability.

-----

Note: I guess the `refactor` part of this PR doesn't really impact performance but I had to break the event_handler file down to debug so might as well push it! 
On that note, everything before d7a9a4a should just fix the bug and include tests, so bug fix doesn't depend on refactor.

And also I tried to retain the core funtionality of event_handler in this refactor, but I did take a different approach to handle events in teh test file under that module to ensure tests reliability.


k0kubun:
> On that note, everything before https://github.com/xremap/xremap/commit/d7a9a4aab39b3021fd52eba8721afee79fd995ed should just fix the bug and include tests, so bug fix doesn't depend on refactor.

Hm. I'll cherry-pick the fix and release it first then. Thank you!

PR #742, opened 2025-09-07
Add support for mapping one key to multiple in modmap

Esper89:
I've implemented a new feature allowing a single key to be mapped to multiple keys in `modmap`. This feature is particularly useful for video games that don't allow one key to be bound to multiple in-game actions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -243,13 +243,15 @@ modmap:
     remap: # Required
       # Replace a key with another
       KEY_XXX1: KEY_YYY # Required
+      # Replace a key with multiple keys (pressed and released simultaneously)
+      KEY_XXX2: [KEY_YYY, KEY_ZZZ]
       # Dispatch different keys depending on whether you hold it or press it alone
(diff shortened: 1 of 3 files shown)


k0kubun:
I'm open to merging it, but I want to make sure I understand the use case.

Is the game's behavior different between `a: [b, c]` in `modmap` with this PR (press B, press C, release B, release C) and the same thing in the `keymap` (press B, release B, press C, release C)?

Esper89:
The benefit of supporting this in `modmap` is that it works properly with modifier keys and keys that need to be held down. For example, if holding Left Shift in a game caused you to crouch and holding X in a game caused you to shoot, adding `Q: [LEFTSHIFT, X]` to `modmap` would allow you to bind both of these actions to the same key, but doing the same in `keymap` wouldn't work properly with Left Shift and would only tap X instead of holding it down.

k0kubun (approved):
:ok_hand: Thanks for the explanation

PR #823, opened 2026-01-16
Secure / Isolated KDE Wayland setup

zine999:
Add documentation and implementation for securely running xremap on KDE Plasma Wayland.

This new guide details a setup where xremap runs as a dedicated system user. This architecture isolates input device access, adhering to the principle of least privilege and preventing malicious user applications from logging keystrokes.

It includes instructions for:

- Creating a dedicated system user with access to the `input` group.
- Implementing a socket-based helper to bridge KWin window info to the isolated xremap process for application-specific remapping.
- Configuring systemd services for both the main xremap instance and the helper.

The setup has been tested on Fedora and works across reboots.

I'm opening this PR for feedback. Note that this also has the changes from #821.

**NOTE:** I'm not a Rust or Python developer, these changes were implemented by AI under guidance and testing. Feel free to close this PR or take over.

### DIFF
--- a/doc/kde-wayland/xremap-tmpfiles.conf
+++ b/doc/kde-wayland/xremap-tmpfiles.conf
@@ -1,0 +1,4 @@
+# Create xremap socket directory
+# Owner: root, Group: root, Mode: 1777 (world-writable with sticky bit)
+# Sticky bit prevents users from deleting each other's files
+d /run/xremap 1777 root root -
--- a/doc/README.md
+++ b/doc/README.md
@@ -8,6 +8,7 @@ You can choose between the following ways to run xremap. They are ordered by the
 - [Running without sudo](running_without_sudo.md)
 - [Running as a user service](running_as_user_service.md)
 - [Running as a system service](running_as_system_service.md)
+- [Running on KDE Wayland as an isolated service](kde-wayland/README.md)
 
 ### Troubleshooting
 
--- a/doc/kde-wayland/README.md
+++ b/doc/kde-wayland/README.md
@@ -1,0 +1,870 @@
+# Xremap Setup Guide for KDE Plasma Wayland
+
+This guide explains how to securely run xremap on KDE Plasma Wayland using a dedicated system user and socket-based helper for application-specific key remapping.
+
+## Table of Contents
+
+1. [Overview and Security Model](#overview-and-security-model)
+2. [Architecture](#architecture)
+3. [Prerequisites](#prerequisites)
+4. [Installation](#installation)
+5. [Configuration](#configuration)
+6. [Verification](#verification)
+7. [Troubleshooting](#troubleshooting)
+8. [Cleanup and Removal](#cleanup-and-removal)
+
+---
+
+## Overview and Security Model
+
+### The Security Problem
+
+When xremap needs to remap keyboard input, it requires access to input devices (`/dev/input/event*`). On Linux, this access is controlled by the `input` group. The security concern is:
+
+- **Dangerous approach**: Add your regular user (e.g., `yourusername`) to the `input` group
+  - **Problem**: *Any application* running as your user can now read keyboard input
+  - **Consequence**: Malicious software, compromised apps, or even buggy programs could log your keystrokes (passwords, chat messages, etc.)
+
+### The Secure Solution
+
+Instead, xremap should run as a **dedicated system user**:
+
+```
+┌─────────────────────────────────────────────────────────────┐
+│   Your User (`yourusername`)                                │
+│   - Runs your applications (browser, terminal, IDE, etc.)   │
+│   - NOT in the `input` group                                │
+│   - Cannot access raw keyboard input                        │
+│   - Even if compromised, cannot read keystrokes             │
+└─────────────────────────────────────────────────────────────┘
+
+┌────────────────────────────────────────────────────────────┐
+│   `xremap` System User                                     │
+│   - Runs ONLY the xremap process                           │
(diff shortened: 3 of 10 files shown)


doc/kde-wayland/xremap-tmpfiles.conf
```diff
@@ -0,0 +1,4 @@
+# Create xremap socket directory
+# Owner: root, Group: root, Mode: 1777 (world-writable with sticky bit)
+# Sticky bit prevents users from deleting each other's files
+d /run/xremap 1777 root root -
```
millerdev: Does this mean that (without a restrictive umask), any user will be able to read any other user's socket at `/run/xremap/kde.sock`?

And with a restrictive umask (that blocks world read/write), the `xremap` process will not have permission to access the socket?

zine999: Gemini's answer:

<details>
Based on the Python code provided, specifically this line:

```python
# Set permissions to allow xremap user to connect
os.chmod(str(self.socket_path), 0o666)
```

Here are the answers to your questions:

### 1. Does this mean any user will be able to read/access the socket?

**Yes.**

The script explicitly sets the socket permissions to `0o666` (Read/Write for Owner, Group, and **Others**).
*   **The Directory:** The `1777` permissions on `/run/xremap` allow users to enter the directory.
*   **The Socket:** The `666` permissions on `kde.sock` allow any process on the system to connect to the socket.

Because the code explicitly forces these permissions, any user on the system can connect to this socket and read the data being sent (which appears to be the currently active window title, class, and app_id). This is a privacy leak (other users can see what applications/windows you are using), though likely not a system-compromise risk unless the script also accepts incoming commands.

### 2. With a restrictive umask, will `xremap` be blocked?

**No.**

The `umask` of the user running the script **will be ignored/overridden** by the code.

Here is the sequence of events in the script:
1.  `self.server_socket.bind(...)`: The socket file is created. At this exact microsecond, the `umask` applies (e.g., creating it as `0700` or `0755`).
2.  `os.chmod(..., 0o666)`: The script immediately changes the permissions to `rw-rw-rw-`.

Therefore, `xremap` (regardless of whether it runs as `root`, `user`, or `nobody`) will **always** be able to access the socket, because the script forces the door wide open for everyone.

### Summary
The author of this script intentionally prioritized ease of configuration over security. By setting `0o666`, they ensure that `xremap` (often running as root) can talk to the KDE helper (running as a user) without needing complex group permissions or ACLs. The trade-off is that any other user on the system can also "listen in" on your window changes.
</details>

I'm open to suggestions on how to modify this if they keep everything working.

zine999:
@hpccc53 I've made your changes requests regarding #821 adding `--log-window-changes` both there and to this PR as well.

> If you wish to move forward with the other functionality. I can give you some suggestions, but it requires some rework.

I'm certainly open to making changes, although note any suggestions you give I'll simply pass along to the AI to do. As long as the functionality remains working and no security issues are re-introduced I've no problems with improving this PR, I'm sure the AI didn't do a perfect job and am happy to keep iterating on it as time allows.

doc/kde-wayland/xremap-tmpfiles.conf
zine999: Also, what is the security concern here? My understanding is that this socket would allow any process to read the window titles and apps that are being accessed by the user, but the user already has permission to do that, so any process running under the user already has that capability.

millerdev: I'm suggesting `xremap-socket` in addition because it automatically switches to the socket of the active user session, which is in a secure location that can only be accessed by that user and xremap.

I would like—and am going to try—to integrate `xremap-socket` functionality into `xremap` so the extra service is not needed, but for now I think it is the most secure option.

zine999:
@millerdev Can your #827 be used with KDE?

hpccc53:
I have released a `bridge` in v0.15.1, which will do the same thing as this PR. The `bridge` makes application-specific remapping work for all desktop environments.  [The documentation is updated on how to use it](https://github.com/xremap/xremap/blob/master/doc/running_as_system_service.md).


Feedback on how it works is appreciated.

HotwheelsSisyphus:
Thanks! Hoping to set this up this weekend... 

> The bridge makes application-specific remapping work for all desktop environments

I looked at the updated docs, but I'm not sure I understand... so just to make sure I get this right:
1. Use the "socket" version of the binary as a system service;
2. use the "kde" version of the binary as a user service, with the --bridge flag and nothing else;
3. will I need to make separate .service files for the two instances? 

EDIT: got it to work! To answer my own question 3, yes, but the .service file for the user service is minimal. I think the user service needs to have started before the system service starts. I haven't yet rebooted the system so I'm not sure how this will behave on startup.
so far, application-specific maps work with no obvious problems. will report back if there are any! (mostly using it for umlaut/diacritics across different applications).

Thanks again for your work, this is really an invaluable tool!

hpccc53:
> will report back if there are any! (mostly using it for umlaut/diacritics across different applications).

Appreciate that, so bugs can be fixed.

> I think the user service needs to have started before the system service starts.
> I haven't yet rebooted the system so I'm not sure how this will behave on startup.

I think it should work. User services are started when you login, and system services
should be started by then, though I don't know if it's guaranteed.

Varstahl:
I can confirm system+bridge works, so this PR is obsolete.

For the ones needing guidance I have drafted a full step by step guide for the installation and configuration of both system service and user services, available at https://github.com/xremap/xremap/discussions/824#discussioncomment-17262209.

PR #947, opened 2026-06-24
Reconnect the COSMIC client when the cosmic-comp IPC drops

ahonnecke:
`CosmicClient::borrow()` only (re)connects when `self.queue` is `None`. On COSMIC, cosmic-comp closes its Wayland IPC periodically (surfaces as `Io error: Broken pipe`). When that happens on a *live* connection, `self.queue` stays `Some`, so every later `queue.roundtrip(state)?` hits the dead socket and returns an error **forever**:

```
Error when fetching app_id: Backend error: Io error: Broken pipe (os error 32)
```

`current_application()` / `current_window()` then return `None` on every keypress, so all `application`/`window`-conditioned remaps silently stop firing until xremap is restarted by hand. The process stays "active", so `Restart=on-failure` never triggers.

**Fix:** in `borrow()`, on a `roundtrip()` error, tear down the dead `queue`/`state` and retry once — the retry re-runs `connect()`. A transient drop is healed on the same lookup; a genuinely-down compositor still returns an error after the single retry, so behavior on a hard failure is unchanged.

**Repro on COSMIC:** run the `cosmic` build with any `application`/`window` keymap, leave it running until cosmic-comp recycles the IPC (or restart cosmic-comp), then observe the remaps stay dead and the journal fills with the Broken-pipe line above. With this change, the next keypress reconnects and remapping resumes.

Cut against `v0.15.6`; `src/client/cosmic_client.rs` is byte-identical on `main`, so it applies cleanly there.

### DIFF
--- a/src/client/cosmic_client.rs
+++ b/src/client/cosmic_client.rs
@@ -83,8 +83,36 @@ impl CosmicClient {
     }
 
     fn borrow<'a>(&'a mut self) -> Result<(&'a mut EventQueue<State>, &'a mut State)> {
-        if self.queue.is_none() {
-            self.connect()?;
+        // cosmic-comp's Wayland IPC drops periodically (surfaces as a "Broken pipe").
+        // The original code only (re)connected when self.queue was None, but a dropped
+        // connection leaves self.queue as Some — so every later roundtrip() hit the dead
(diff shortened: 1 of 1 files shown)


hpccc53:
Thanks. I think it makes sense.

